PR #1, opened 2016-07-18
Initial commit

zheins:
@n1zea144 

Any comments on structure/style/implementation? I also asked Onur to take a look at the Annotation class.

For now I left the option to specify isoform overrides in the properties file (current way our pipelines does it), but I feel like it would be better to have as a command line option. Do you agree?


zheins:
Opened new PR

PR #2, opened 2016-07-29
Initial commit

zheins:
@n1zea144 

Should I change the name of the main class and things to remove references to GenomeNexus since we made the annotation an interface that could technically plug into anything?


n1zea144:
Yes, you could just refer to things as annotation pipeline or something to that effect.  AnnotatorImpl could be renamed to GenomeNexusImpl.

zheins:
@n1zea144 updated PR - no longer has any genome nexus references in the code. We might want to change the name of this repo as well. I also added the option to keep hugo symbols from annotation if desired.

annotationPipeline/src/main/resources/application.properties.EXAMPLE
```diff
@@ -0,0 +1,6 @@
+spring.batch.job.enabled=false
+genomenexus.base_url=
+genomenexus.hgvs=variant_annotation/hgvs/
+genomenexus.isoform_override=variant_annotation/isoform_override/
```
n1zea144: Should this be a runtime argument?  Do we want to have to repackage/launch the pipeline if we want to change isoform files (say between MSK and Uniprot?)

annotator/src/main/java/org/cbioportal/annotator/Annotator.java
```diff
@@ -0,0 +1,47 @@
...
+import org.cbioportal.models.MutationRecord;
+
+/**
+ *
+ * @author heinsz
+ */
+
+public interface Annotator {
+
+    Annotator annotator();
```
n1zea144: Whats this?

annotator/src/main/java/org/cbioportal/models/AnnotatedRecord.java
```diff
@@ -0,0 +1,265 @@
...
+public class AnnotatedRecord extends MutationRecord{
+    protected String hgvsc;
+    protected String hgvsp;
+    protected String hgvspShort;
+    protected String transcriptId;
+    protected String refSeq;
+    protected String proteinPosStart;
+    protected String proteinPosEnd;
+    protected String codonChange;
+    protected static Map<String, String> additionalProperties = new LinkedHashMap<String, String>();
```
n1zea144: Do you need this if you already have one in MutationRecord.java?

annotator/src/main/java/org/cbioportal/models/MutationRecord.java
```diff
@@ -0,0 +1,443 @@
...
+    protected String sequencer;
+    protected String tumorSampleUUID;
+    protected String matchedNormSampleUUID;
+    protected String tRefCount;
+    protected String tAltCount;
+    protected String nRefCount;
+    protected String nAltCount;
+    protected Map<String, String> additionalProperties = new LinkedHashMap<>();
+
+    public MutationRecord() {}
```
n1zea144: Are you ever working with an instance of this class?  Should we add a constructor w/arguments?

annotator/src/main/java/org/cbioportal/models/AnnotatedRecord.java
```diff
@@ -0,0 +1,265 @@
...
+        String hgvsp,
+        String hgvspShort,
+        String transcriptId,
+        String refSeq,
+        String proteinPosStart,
+        String proteinPosEnd,
+        String codonChange,
+        Map additionalProperties
+    ) {        
+        this.hugoSymbol = hugoSymbol;
```
n1zea144: I think its cleaner if you add a call to super(hugoSymbol, entrezGeneId, ....)

annotator/src/main/java/org/cbioportal/models/AnnotatedRecord.java
```diff
@@ -0,0 +1,265 @@
...
+        return this.codonChange;
+    }
+    
+    public void setCodons(String codonChange) {
+        this.codonChange = codonChange;
+    }        
+    
+    public static List<String> getFieldNames() {
+        List<String> fieldNames = new ArrayList<>();
+
```
n1zea144: Should you call super() here?

annotator/src/main/java/org/cbioportal/models/GenomeNexusAnnotationResponse.java
```diff
@@ -0,0 +1,329 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+package org.cbioportal.models;
+
+import java.util.ArrayList;
```
n1zea144: In general, if I have more than one class coming from the same package, I use a wildcard '*'.

annotator/src/main/java/org/cbioportal/models/GenomeNexusAnnotationResponse.java
```diff
@@ -0,0 +1,329 @@
+/*
```
n1zea144: Can you fix the indentation in this class?

annotator/src/main/java/org/cbioportal/models/GenomeNexusIsoformOverridesResponse.java
```diff
@@ -0,0 +1,182 @@
+/*
```
n1zea144: Can you fix the annotation in this class?

annotator/src/main/java/org/cbioportal/models/TranscriptConsequence.java
```diff
@@ -0,0 +1,452 @@
+/*
```
n1zea144: Can you fix the indentation in this class?

annotationPipeline/pom.xml
```diff
@@ -0,0 +1,41 @@
...
+    <version>0.1.0</version>
+  </parent>
+
+  <dependencies>
+      <dependency>
+          <groupId>org.cbioportal.annotator</groupId>
+          <artifactId>annotator</artifactId>
+          <version>0.1.0</version>
+      </dependency>
+    <dependency>
```
n1zea144: Do you still need the spring-web dependency?

annotationPipeline/src/main/java/org/cbioportal/annotation/AnnotationPipeline.java
```diff
@@ -0,0 +1,96 @@
...
+@SpringBootApplication
+public class AnnotationPipeline
+{
+
+    private static Options getOptions(String[] args)
+    {
+        Options gnuOptions = new Options();
+        gnuOptions.addOption("h", "help", false, "shows this help document and quits.")
+            .addOption("f", "filename", true, "Mutation filename")
+            .addOption("d", "directory", true, "Output directory")
```
n1zea144: I think directory should be an output filename (which can include the path).  This simplifies the code in the MutationRecordWriter and you don't need to capture a genomenexus.output_filename property (which is not very flexible).

annotationPipeline/src/main/java/org/cbioportal/annotation/pipeline/BatchConfiguration.java
```diff
@@ -0,0 +1,106 @@
...
+            .build();
+    }   
+    
+    
+
+    @Bean
+    public Step step()
+    {
+        return stepBuilderFactory.get("step")
+            .<AnnotatedRecord, String> chunk(10)
```
n1zea144: Can we make chunk an application property?

annotationPipeline/src/main/java/org/cbioportal/annotation/pipeline/MutationRecordReader.java
```diff
@@ -0,0 +1,157 @@
...
+                    break;
+                }
+            }            
+            reader.close();
+        }        
+        catch (Exception e) {
+            throw new ItemStreamException(e);
+        }
+        
+        // Add comments to the config for the writer to access later
```
n1zea144: I like Angelica's approach - to use the job context not the Batch Configuration to share data across the steps.

annotationPipeline/src/main/resources/application.properties.EXAMPLE
zheins: I think it's probably better as a runtime argument. I did it this way just because the current way we set the isoforms annotation is through the properties file.

zheins:
@n1zea144 All the proposed changes have been made.

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -0,0 +1,614 @@
...
+ * 
+ * Use GenomeNexus to annotate mutation records
+ * 
+ */
+
+@Configuration
+public class GenomeNexusImpl implements Annotator {
+    
+    @Value("${genomenexus.base_url}")
+    private  String baseUrl;
```
n1zea144: I think you should get rid of this base URL from here and the application.properties file.  If you should just fully form the URLs in the application.properties file and not do the combining in java code.  Maybe its possible to reference #{genomenexus.base_url} in the application.properties class so you don't have to rewrite the base_url in multiple locations, but I think thats a cleaner solution.

zheins:
@n1zea144 didn't see a way to reference a property within the property file, so I just made the initial change suggested of requiring fully formed URLs in application.properties.

annotationPipeline/src/main/resources/application.properties.EXAMPLE
n1zea144: I think we should make this switch.  This way we can use the same compiled codebase in our cronjob to annotate the MSK-IMPACT data (MSK isoforms) for the MSK portal or public studies (UniProt) for the public portal.

annotator/src/main/java/org/cbioportal/models/MutationRecord.java
```diff
@@ -0,0 +1,495 @@
...
+    
+    public Map<String, String> getAdditionalProperties() {
+        return this.additionalProperties;
+    }
+    
+    public void setAdditionalProperties(Map<String, String> additionalProperties) {
+        this.additionalProperties = additionalProperties;
+    } 
+    
+    public List<String> getFieldNames() {
```
n1zea144: This is minor, but whats the point of these methods (same one in AnnotatedRecord)?  Why not hardcode a member list in a static function instead of recomputing for each call and just have getHeader() as part of the "public" api?

annotator/src/main/java/org/cbioportal/models/GenomeNexusIsoformOverridesResponse.java
```diff
@@ -0,0 +1,181 @@
...
+    public Map<String, Object> getAdditionalProperties() {
+        return this.additionalProperties;
+    }
+
+    @JsonAnySetter
+    public void setAdditionalProperty(String name, Object value) {
+        this.additionalProperties.put(name, value);
+    }
+
+}
```
n1zea144: Please add a newline here.

n1zea144:
I think we need to add at least one unit test to the project.  You can check in an pre and post annotated MAF.  The test can create a new annotated MAF from the checked in pre- and diff it with the checked-in post annotated MAF.

zheins:
@n1zea144 added integration testing. Ready for review.

annotator/src/main/java/org/cbioportal/models/MutationRecord.java
n1zea144: Did you see this comment?

n1zea144: Nevermind, I see some updates.

annotationPipeline/src/test/resources/application-test.properties.EXAMPLE
```diff
@@ -0,0 +1,4 @@
+spring.batch.job.enabled=false
+genomenexus.hgvs=http://<HOST>:5558/variant_annotation/hgvs/
```
n1zea144: Please remove the port nums from here.

PR #6, opened 2016-11-21
Call for annotation specifying isoform

zheins:
Simplifies the logic for selecting the "canonical" transcript by specifying it in the annotation call to genome-nexus. Also improves performance by not making two calls per variant.

Reviewers:
@n1zea144, Onur


onursumer:
@zheins: looks good to me

PR #7, opened 2016-12-07
Check for incorrect missense mutation classifications

zheins:
If variant is `protein_altering_variant` or `coding_sequence_variant`, check if the variant type is DEL or INS, and set classification to in frame insertion/deletion accordingly. 

@n1zea144 @jjgao

### DIFF
--- a/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
+++ b/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
@@ -649,8 +649,8 @@ public class GenomeNexusImpl implements Annotator {
             variantMap.put("inframe_insertion",             "In_Frame_Ins");
             variantMap.put("inframe_deletion",              "In_Frame_Del");    
             variantMap.put("missense_variant",              "Missense_Mutation");
-            variantMap.put("protein_altering_variant",      "Missense_Mutation"); // Not sure if this is correct
-            variantMap.put("coding_sequence_variant",       "Missense_Mutation");
+            variantMap.put("protein_altering_variant",      "Missense_Mutation"); // Not always correct, code below to handle exceptions
+            variantMap.put("coding_sequence_variant",       "Missense_Mutation"); // Not always correct, code below to handle exceptions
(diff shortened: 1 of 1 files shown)


annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -649,8 +649,8 @@ private String getVariantClassificationFromMap(String variant) {
             variantMap.put("inframe_insertion",             "In_Frame_Ins");
             variantMap.put("inframe_deletion",              "In_Frame_Del");    
             variantMap.put("missense_variant",              "Missense_Mutation");
-            variantMap.put("protein_altering_variant",      "Missense_Mutation"); // Not sure if this is correct
-            variantMap.put("coding_sequence_variant",       "Missense_Mutation");
+            variantMap.put("protein_altering_variant",      "Missense_Mutation"); // Not correct, code below to handle it
```
onursumer: Should we change the comment to "Not always correct, code below to handle exceptions"?

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -684,7 +684,14 @@ private String getVariantClassificationFromMap(String variant) {
             variantMap.put("feature_elongation",            "Targeted_Region"); 
             variantMap.put("feature_truncation",            "Targeted_Region"); 
         }
-        
+        if (variant.toLowerCase().equals("protein_altering_variant") || variant.toLowerCase().equals("coding_sequence_variant")) {
+            if (resolveVariantType().equals("DEL")) {
+                return "In_Frame_Del";
+            }
+            else if (resolveVariantType().equals("INS")) {
```
onursumer: We are calling the `resolveVariantType` method twice here. If it is a costly operation, it might be better to call it once and assign it to a variable.

zheins: It's not costly, but should just be a variable.

PR #13, opened 2017-01-25
Added logic for handling inframe_insertion canonical transcripts

angelicaochoa:
1. Added logic for handling inframe_insertion and deletion canonical transcripts
Some inframe_insertion cases were being missed by genome nexus when resolving protein change.

2. Added code to correctly resolve genome assembly name and variant classification
If genome nexus could not resolve a variant then these fields would be left empty instead of retaining what was in the input MAF file

3. Fixed how `MutationRecordProcessor` generates a line record to pass to the `MutationRecordWriter`
Extra tabs were being inserted at the end of the line, creating one extra field in the row relative to the MAF header

4. Added logging to `MutationRecordReader` to alert whether a variant could not be annotated

### DIFF
--- a/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
+++ b/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
@@ -86,6 +86,8 @@ public class GenomeNexusImpl implements Annotator {
         {"Leu", "L"}, {"Lys", "K"}, {"Met", "M"}, {"Phe", "F"}, {"Pro", "P"}, {"Ser", "S"},
         {"Thr", "T"}, {"Trp", "W"}, {"Tyr", "Y"}, {"Val", "V"}, {"Xxx", "X"}, {"Ter", "*"}
     };
+    private Set<String> spliceSiteVariants = new HashSet<String>(Arrays.asList(
+            "splice_acceptor_variant", "splice_donor_variant", "splice_region_variant"));
 
     private Map<String, String> variantMap = new HashMap<String, String>();
     private List<String> hgvspNullClassifications = new ArrayList<>();
@@ -180,7 +182,7 @@ public class GenomeNexusImpl implements Annotator {
         AnnotatedRecord annotatedRecord= new AnnotatedRecord(resolveHugoSymbol(replaceHugo),
                 mRecord.getEntrez_Gene_Id(),
                 mRecord.getCenter(),
-                gnResponse.getAssemblyName(),
+                resolveAssemblyName(),
                 resolveChromosome(),
                 resolveStart(),
                 resolveEnd(),
@@ -254,6 +256,10 @@ public class GenomeNexusImpl implements Annotator {
         return mRecord.getChromosome();
     }
 
+    private String resolveAssemblyName() {
+        return (gnResponse.getAssemblyName() == null) ? mRecord.getNCBI_Build() : gnResponse.getAssemblyName();
+    }
+    
     private String resolveStart() {
         try {
             if (gnResponse.getStart() != null) {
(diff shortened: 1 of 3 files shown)


annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -409,7 +409,7 @@ private String resolveHgvspShort() {
                     }
                 }
             }
-            else if (canonicalTranscript.getHgvsc() != null) {
+            else if (canonicalTranscript.getHgvsc() != null && !canonicalTranscript.getConsequenceTerms().get(0).equals("inframe_insertion")) {
```
zheins: What about inframe deletions or missense? Maybe change this logic to look for the cases that are being used inside. Add those here instead so that it only goes in that block for splice site stuff.

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -427,21 +427,34 @@ else if (canonicalTranscript.getHgvsc() != null) {
...
-                    else {
-                        hgvsp += aaParts[1];
-                    }
+                hgvsp = resolveHgvspShortFromAAs();
+            }
+        }
+        return hgvsp;
+    }
+    
+    private String resolveHgvspShortFromAAs() {
```
zheins: I like having this pulled out as a method.

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -427,21 +427,34 @@ else if (canonicalTranscript.getHgvsc() != null) {
...
+        String hgvsp = "";
+        try {
+            String[] aaParts = canonicalTranscript.getAminoAcids().split("/");
+            if (canonicalTranscript.getConsequenceTerms() != null && canonicalTranscript.getConsequenceTerms().get(0).equals("inframe_insertion")) {
```
zheins: Probably should also do this for deletions

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -254,6 +256,10 @@ private String resolveChromosome() {
         return mRecord.getChromosome();
     }
 
+    private String resolveAssemblyName() {
+        return (gnResponse.getAssemblyName() == null)?mRecord.getNCBI_Build():gnResponse.getAssemblyName();
```
zheins: Can you fix the spacing here?

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -319,10 +325,8 @@ private String resolveVariantClassification() {
         if (gnResponse.getAlleleString() != null) {
             alleles = gnResponse.getAlleleString().split("/", -1);
         }
-        if(alleles != null) {
-            if (alleles.length == 2) {
-                variantType = getVariantType(alleles[0], alleles[1]);
-            }
+        if(alleles != null && alleles.length == 2) {
```
zheins: good fix

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -409,39 +413,50 @@ private String resolveHgvspShort() {
...
+            }
+            else {
+                hgvsp = aaParts[0] + canonicalTranscript.getProteinStart();
+                if (canonicalTranscript.getConsequenceTerms() != null && canonicalTranscript.getConsequenceTerms().get(0).equals("frameshift_variant")) {
+                    hgvsp += "fs";
                 }
-                catch (NullPointerException e) {
-                    log.debug("Failed to salvage HGVSp_Short from protein start, amino acids, and consequence terms");
+                else {
+                    hgvsp += aaParts[1];
```
zheins: Is this proper behavior? I think better to just accept that it didn't annotate properly and keep the log? Maybe upgrade the log from debug level to error level.

zheins: Sorry, misread this. Looks good.

PR #14, opened 2017-02-15
Genome Nexus pipeline enhancements

angelicaochoa:
Added some more detailed logging when a record fails to annotate
* this helps indicate whether a variant failed to annotate because there's something wrong with the data itself or if it failed to annotate because it's a non-coding region, mitochondrial variant, etc.

Added logging to reader that summarizes total records that failed to annotate 
* also includes breakdown of how many variants failed to annotate because of HGVSp null variant classifications or because they are mitochondrial variants

Standardized job parameters configuration for `replace` parameter

### DIFF
--- a/annotator/src/main/java/org/cbioportal/annotator/Annotator.java
+++ b/annotator/src/main/java/org/cbioportal/annotator/Annotator.java
@@ -42,6 +42,7 @@ import java.util.Map;
 public interface Annotator {
      
     AnnotatedRecord annotateRecord(MutationRecord record, boolean replaceHugo, String isoformOverride, boolean reannotate);
-    MutationRecord createRecord(Map<String, String> mafLine) throws Exception; 
+    MutationRecord createRecord(Map<String, String> mafLine) throws Exception;
+    boolean isHgvspNullClassifications(String variantClassification);
     
 }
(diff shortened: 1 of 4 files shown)


annotator/src/main/java/org/cbioportal/annotator/Annotator.java
```diff
@@ -42,6 +42,7 @@
 public interface Annotator {
      
     AnnotatedRecord annotateRecord(MutationRecord record, boolean replaceHugo, String isoformOverride, boolean reannotate);
-    MutationRecord createRecord(Map<String, String> mafLine) throws Exception; 
+    MutationRecord createRecord(Map<String, String> mafLine) throws Exception;
```
zheins: What changed on this line?

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -766,6 +766,11 @@ private String getTumorSeqAllele(MutationRecord record) {
             return tumorSeqAllele = record.getTumor_Seq_Allele1();
         }
     }
+    
+    @Override
+    public boolean isHgvspNullClassifications(String variantClassification) {
```
zheins: Can you explain what this is? Where does `hgvspNullClassifications` get set up

PR #15, opened 2017-02-17
Fix for checking when a record is already annotated

zheins:
Check whether the variant has previously been annotated in an untranslated region before attempting to annotate.

@n1zea144 @sheridancbio

### DIFF
--- a/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
+++ b/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
@@ -90,7 +90,7 @@ public class GenomeNexusImpl implements Annotator {
             "splice_acceptor_variant", "splice_donor_variant", "splice_region_variant"));
 
     private Map<String, String> variantMap = new HashMap<String, String>();
-    private List<String> hgvspNullClassifications = new ArrayList<>();
+    private static List<String> hgvspNullClassifications = initNullClassifications();
 
     private Pattern cDnaExtractor = Pattern.compile(".*[cn].-?\\*?(\\d+).*");
     private String variantType;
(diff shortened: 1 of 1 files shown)


annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -112,7 +121,7 @@ public AnnotatedRecord annotateRecord(MutationRecord record, boolean replaceHugo
 
         //check if record already is annotated
         Map<String, String> additionalProperties = mRecord.getAdditionalProperties();
-        if(additionalProperties.containsKey("HGVSp_Short") && !reannotate && !additionalProperties.get("HGVSp_Short").isEmpty()) {
+        if(additionalProperties.containsKey("HGVSp_Short") && !reannotate && !additionalProperties.get("HGVSp_Short").isEmpty() || hgvspNullClassifications.contains(mRecord.getVariant_Classification())) {
```
n1zea144: This doesn't change the logic, but I think it reads better if you have:

if (additionalProperties.containsKey("HGVSp_Short") && !additionalProperties.get("HGVSp_Short").isEmpty() && !reannotate...

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -90,7 +90,16 @@
             "splice_acceptor_variant", "splice_donor_variant", "splice_region_variant"));
 
     private Map<String, String> variantMap = new HashMap<String, String>();
-    private List<String> hgvspNullClassifications = new ArrayList<>();
+    private static List<String> hgvspNullClassifications = new ArrayList<>();
```
n1zea144: This is a nice pattern for initializing static members:

private static List<String> hgvspNullClassifications = initNullClassifications();
private static List<String> initNullClassifications()
{
...
}

sheridancbio: since this can be an immutable list, it can be initialized like this instead:
List<String> strings = Arrays.asList("foo", "bar", "baz");
see  http://stackoverflow.com/questions/1005073/initialization-of-an-arraylist-in-one-line

If the list was longer, I might consider making it a hashset .. that way lookup into the list would be ~ order(1) rather than order(n). But the overhead of hashing is probably more costly when the list size is under 10

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -105,14 +105,19 @@ public GenomeNexusImpl annotator(String hgvsServiceUrl) {
         this.hgvsServiceUrl = hgvsServiceUrl;
         return this;
     }
+    
+    private boolean annotationNeeded(MutationRecord record) {
+        Map<String, String> additionalProperties = record.getAdditionalProperties();
+        return additionalProperties.containsKey("HGVSp_Short") && additionalProperties.get("HGVSp_Short").isEmpty() && !hgvspNullClassifications.contains(record.getVariant_Classification());
```
sheridancbio: Sorry to continue to make this complicated, but doesn't this code miss the case where additionalProperties does not contain the key "HGVSp_Short" ?  if it does not, then this method will return false even though annotationNeeded() should be true.

Since we are in a separate method, we don't need to string all these conditions into a single logical expression. Here is my guess as to the logic:

Map<String, String> additionalProperties = record.getAdditionalProperties();
if (!additionalProperties.containsKey("HGVSp_Short") {
return true;
}
return additionalProperties.get("HGVSp_Short").isEmpty() && !hgvspNullClassifications.contains(record.getVariant_Classification());

in other words "if the key was never created, we always have to annotate. If the value of the key is empty we *may* need to annotate, unless the VariantClassification has been set to a non-protein-change type of mutation."

I could be wrong about this ... I think @n1zea144 should weigh in. We can also revert back to the original approach if we have expended too much time on this effort.

zheins: @sheridancbio Your logic here is correct. Great catch!

PR #16, opened 2017-03-08
Fix issue with start pos on insertions

zheins:
Thanks @ckandoth for solving this.

@angelicaochoa could you review?


angelicaochoa:
This PR handles cases where the ref allele is an empty string (like VCF insertion events), which is handled by decrementing the given start position by 1 before trying to annotate the variant.

PR #17, opened 2017-03-10
Adding consequence column

zheins:
Adding a field in the annotated record for VEP consequence_terms, called Consequences.

@angelicaochoa @sheridancbio

### DIFF
--- a/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
+++ b/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
@@ -168,6 +168,7 @@ public class GenomeNexusImpl implements Annotator {
                 additionalProperties.get("Protein_Position") != null ? additionalProperties.get("Protein_Position") : "",
                 additionalProperties.get("Codons") != null ? additionalProperties.get("Codons") : "",
                 additionalProperties.get("Hotspot") != null ? additionalProperties.get("Hotspot") : "",
+                additionalProperties.get("Consequence") != null ? additionalProperties.get("Consequence") : "",
                 additionalProperties);
         }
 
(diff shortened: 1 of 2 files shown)


annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -168,6 +168,7 @@ public AnnotatedRecord annotateRecord(MutationRecord record, boolean replaceHugo
                 additionalProperties.get("Protein_Position") != null ? additionalProperties.get("Protein_Position") : "",
                 additionalProperties.get("Codons") != null ? additionalProperties.get("Codons") : "",
                 additionalProperties.get("Hotspot") != null ? additionalProperties.get("Hotspot") : "",
+                additionalProperties.get("Consequence") != null ? additionalProperties.get("Consequcne") : "",
```
angelicaochoa: typo here

annotator/src/main/java/org/cbioportal/models/AnnotatedRecord.java
```diff
@@ -58,7 +59,8 @@
         HEADER.add("Protein_position");
         HEADER.add("Codons");        
         HEADER.add("Hotspot");
-    }       
+        HEADER.add(HEADER.indexOf("Variant_Classification"), "Consequence");
```
angelicaochoa: We want to replace the column `Variant_Classification` with `Consequence`? I thought we just wanted to add the `Consequence` column?

zheins: Add does not replace the element, it sticks in at that position and moves all other elements back one index.

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
zheins: thanks!

annotator/src/main/java/org/cbioportal/models/AnnotatedRecord.java
angelicaochoa: thanks :)

PR #18, opened 2017-03-13
Added url getter for mutation record

zheins:


annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -252,6 +252,17 @@ public MutationRecord createRecord(Map<String, String> mafLine) throws Exception
...
 
+     @Override
+    public boolean isHgvspNullClassifications(String variantClassification) {
+        return hgvspNullClassifications.contains(variantClassification);
+    }
+
+    @Override
+    public String getUrlForRecord(MutationRecord record, String isoformOverridesSource) {
+        String hgvsNotation = convertToHgvs(record);
+        return hgvsServiceUrl + hgvsNotation + "?" + isoformQueryParameter + "=" + isoformOverridesSource + "&"  + hotspotParameter + "=summary";
```
angelicaochoa: let's also use this method for the `annotateRecord()` method [here](https://github.com/zheins/genome-nexus-annotation-client/blob/e97c657fc34deb16f67cc4c35017779b039f462c/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java#L184)

n1zea144: @angelicaochoa Good idea.

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -782,11 +793,6 @@ private String getTumorSeqAllele(MutationRecord record) {
         }
     }
 
-    @Override
-    public boolean isHgvspNullClassifications(String variantClassification) {
-        return hgvspNullClassifications.contains(variantClassification);
-    }
-
     private static List<String> initNullClassifications() {
```
n1zea144: I would move this up to where it is used.  More importantly, we should figure out a better way to keep this list in sync/shareable with [MutationFilter.java](https://github.com/cBioPortal/cbioportal/blob/master/core/src/main/java/org/mskcc/cbio/portal/scripts/MutationFilter.java)

zheins: As it is an interface method, I moved it next to the others. Where would you suggest it be? It's not used in this class, it's basically a way to interact with the list of null classifications described here.

zheins:
@n1zea144 @angelicaochoa 

getUrlForRecord is now being used throughout.

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -172,16 +172,11 @@ public AnnotatedRecord annotateRecord(MutationRecord record, boolean replaceHugo
...
-        String hgvsNotation = convertToHgvs(record);
-
         // make the rest call to genome nexus
         RestTemplate restTemplate = new RestTemplate();
         HttpEntity<LinkedMultiValueMap<String, Object>> requestEntity = getRequestEntity();
 
-        log.debug("Annotating: " + hgvsNotation + " from sample " + record.getTumor_Sample_Barcode());
-
-        ResponseEntity<GenomeNexusAnnotationResponse[]> responseEntity = restTemplate.exchange(hgvsServiceUrl + hgvsNotation + "?" + isoformQueryParameter + "=" + isoformOverridesSource + "&" + hotspotParameter + "=summary", HttpMethod.GET, requestEntity, GenomeNexusAnnotationResponse[].class);
+        ResponseEntity<GenomeNexusAnnotationResponse[]> responseEntity = restTemplate.exchange(getUrlForRecord(record, isoformOverridesSource), HttpMethod.GET, requestEntity, GenomeNexusAnnotationResponse[].class);
```
sheridancbio: good change - the inline construction of the URL made it hard to see the arguments to exchange()

sheridancbio:
Merged now, but if relocation of isHgvspNullClassifications() to another class is desired it can be done in a separate PR

@n1zea144

PR #19, opened 2017-03-20
Do not use static header in mut record

zheins:
Static header causes an issue where after an annotated record was instantiated, all mutation records will also have extra fields in the header they shouldn't have. Use initHeader design instead, and add items to the header in the annotation record constructors.

@sheridancbio @angelicaochoa


sheridancbio:
The error detected here was when incoming mutation records contains a field header which matched in the extra fields present in AnnotatedRecord --- these extra headers fields *were* present in instances of MutationRecord once the class AnnotatedRecord was referenced/initialized, making them candidates for modeling, even though no setter for the field existed in the model.

sheridancbio:
Waiting for merge to allow time for @angelicaochoa to review / suggest changes.

PR #24, opened 2017-04-06
Added http 500 exception handling in genome nexus pipeline

angelicaochoa:


annotationPipeline/src/main/java/org/cbioportal/annotation/pipeline/MutationRecordReader.java
```diff
@@ -106,25 +107,31 @@ public void handleLine(String line) {
...
-                else if (annotatedRecord.getChromosome().equals("M")) {
-                    failedMitochondrialAnnotations++;
-                    message += " - Cannot annotate mitochondrial variants at this time.";
-                }
-                failedAnnotations++;
-                LOG.info(message);
-            }            
-            annotatedRecords.add(annotatedRecord);
-            header.addAll(annotatedRecord.getHeaderWithAdditionalFields());
+            try {
```
zheins: Why does this try need to be around all this code? Couldn't it just be around the actual call to genome nexus at line 111?

PR #33, opened 2017-07-19
Added support for gene external references

angelicaochoa:
If user provides option to replace hugo symbols and entrez ids then these are updated by what's provided from the annotator.

### DIFF
--- a/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
+++ b/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
@@ -36,6 +36,7 @@ import org.cbioportal.models.AnnotatedRecord;
 import org.cbioportal.models.GenomeNexusAnnotationResponse;
 import org.cbioportal.models.MutationRecord;
 import org.cbioportal.models.GenomeNexusIsoformOverridesResponse;
+import org.cbioportal.models.GeneXref;
 import org.cbioportal.annotator.Annotator;
 
 import java.util.*;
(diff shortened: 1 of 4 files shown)


annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -507,6 +510,25 @@ private String resolveConsequence() {
...
     }
+    
+    private String resolveEntrezGeneId(boolean replace) {
+        if (!replace) {
+            return mRecord.getEntrez_Gene_Id();
+        }
+        // make the rest call to ensembl server for gene external refs
+        RestTemplate restTemplate = new RestTemplate();
+        HttpEntity<LinkedMultiValueMap<String, Object>> requestEntity = getRequestEntity();        
+        ResponseEntity<GeneXref[]> responseEntity = restTemplate.exchange(geneXrefsServiceUrl + canonicalTranscript.getGeneId(), HttpMethod.GET, requestEntity, GeneXref[].class);
```
sheridancbio: I wonder about cases where a gene used to have multiple transcripts, but then the entrez ID was deprecated and split into two different genes ... one corresponding to one set of transcripts and one corresponding to a different set of transcripts.  In that case the two new genes may each have a canonicalTranscript.

When genomeNexus is called, is the transcript identified by the caller? Otherwise, can't it be difficult to choose between two such "split" genes?

zheins: I'm curious about this as well. If we do xrefs by a transcript ID, would that be safer? You should then for sure only get one entrez id for sure as opposed to many potential ones. Unless I'm thinking about this wrong.

angelicaochoa: @sheridancbio @zheins right now we are resolving which canonical transcript to use...so this is using the ensembl gene accession (`gene_id`) from the canonical transcript we get/use

angelicaochoa: We also can only get this gene external reference info by gene id. We would need much more info to do the same thing by transcript id, which takes into consideration the CDS coordinates and chromosome, if I recall correctly

PR #34, opened 2017-07-31
Check for null canonical transcripts

zheins:
Prevents NPE.

Signed-off-by: Zack Heins <zackheins@gmail.com>

### DIFF
--- a/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
+++ b/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
@@ -213,7 +213,7 @@ public class GenomeNexusImpl implements Annotator {
     }
 
     private String resolveHugoSymbol(boolean replace) {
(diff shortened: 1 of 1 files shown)


annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -512,7 +512,7 @@ private String resolveConsequence() {
     }
     
     private String resolveEntrezGeneId(boolean replace) {
-        if (!replace) {
+        if (!replace || canonicalTranscript == null || canonicalTranscript.getGeneSymbol() == null || canonicalTranscript.getGeneSymbol().trim().length() == 0) {
```
angelicaochoa: we want to check whether the `canonicalTranscript .geneId()` is null or empty here, not gene symbol (that's the ensembl gene accession that the api uses)

zheins: good catch

PR #37, opened 2017-08-02
Make all getters/setters all caps

zheins:
All usages of this package will need to always do comparisons using toUpperCase on the field mappers.

Signed-off-by: zheins <zackheins@gmail.com>

### DIFF
--- a/annotationPipeline/src/main/java/org/cbioportal/annotation/pipeline/MutationFieldSetMapper.java
+++ b/annotationPipeline/src/main/java/org/cbioportal/annotation/pipeline/MutationFieldSetMapper.java
@@ -32,36 +32,67 @@
 
 package org.cbioportal.annotation.pipeline;
 
-import java.util.*;
 import org.cbioportal.models.MutationRecord;
+import java.util.*;
 import org.springframework.batch.item.file.mapping.FieldSetMapper;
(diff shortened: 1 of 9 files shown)


annotationPipeline/src/main/java/org/cbioportal/annotation/pipeline/MutationFieldSetMapper.java
```diff
@@ -32,36 +32,67 @@
...
+        
+        // Here we make sure that the field is in our intersection
+        Set<String> fields = fieldSetNames.stream()
+                .filter(line -> intersection.contains(line.toUpperCase()))
+                .collect(Collectors.toSet());
+        
+        // In this we actually set the value in our bean map
+        fields.stream().forEach(column -> 
+            {
+                recordBeanMap.put(column.toUpperCase(),Strings.isNullOrEmpty(fs.readRawString(column)) ? "" : fs.readRawString(column));
```
angelicaochoa: Just something to keep in mind - reading the raw string will also include trailing white spaces. May want to trim just in case. Same [here](https://github.com/cBioPortal/genome-nexus-annotation-pipeline/pull/37/files?diff=split#diff-1178360edd668d8dd9cd51e477e1a623R94)

PR #38, opened 2017-08-09
Use highest priority consequence to pick out of canonical transcripts

zheins:
We should be picking highest priority consequence out of the canonicals, even if it isn't the most severe.

Signed-off-by: zheins <zackheins@gmail.com>

### DIFF
--- a/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
+++ b/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
@@ -654,16 +654,27 @@ public class GenomeNexusImpl implements Annotator {
     }
 
     private TranscriptConsequence transcriptWithMostSevereConsequence(List<TranscriptConsequence> transcripts, String mostSevereConsequence) {
+        Integer highestPriority = Integer.MAX_VALUE;
+        TranscriptConsequence highestPriorityTranscript = null;
         for (TranscriptConsequence transcript : transcripts) {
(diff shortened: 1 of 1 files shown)


annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -654,16 +654,27 @@ else if (transcripts.size() > 1) {
...
+        TranscriptConsequence highestPriorityTranscript = null;
         for (TranscriptConsequence transcript : transcripts) {
             List<String> consequenceTerms = transcript.getConsequenceTerms();
             for (String consequenceTerm : consequenceTerms) {
+                if (effectPriority.getOrDefault(consequenceTerm.toLowerCase(), Integer.MAX_VALUE) < highestPriority) {
+                    highestPriorityTranscript = transcript;
+                    highestPriority = effectPriority.getOrDefault(consequenceTerm.toLowerCase(), Integer.MAX_VALUE);
+                }
                 if (consequenceTerm.trim().equals(mostSevereConsequence)) {
                     return transcript;
```
angelicaochoa: Wouldn't this always return the most severe consequence if it sees a match and basically override whether we are searching for the most severe consequence?

zheins: If it matches the most_severe_consequence, it returns that. If none of the transcripts match, it returns the transcript with the highest priority consequence instead of just returning the first one.

PR #40, opened 2017-08-22
Adding equals and hashcode for identical variant check

DixitPatel:
This change implements equals and hashcode method to compare two mutation records and check if they are identical variants or not.

### DIFF
--- a/annotator/pom.xml
+++ b/annotator/pom.xml
@@ -32,11 +32,6 @@
       <artifactId>commons-cli</artifactId>
       <version>1.3</version>
     </dependency>
-    <dependency>
-      <groupId>org.cbio</groupId>
-      <artifactId>gdcpipeline</artifactId>
-      <version>0.0.1-SNAPSHOT</version>
-    </dependency>
   </dependencies>
 </project>
   
--- a/annotator/pom.xml
+++ b/annotator/pom.xml
@@ -32,6 +32,11 @@
       <artifactId>commons-cli</artifactId>
       <version>1.3</version>
     </dependency>
+    <dependency>
+      <groupId>org.cbio</groupId>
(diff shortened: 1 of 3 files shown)


annotator/pom.xml
```diff
@@ -32,6 +32,11 @@
       <artifactId>commons-cli</artifactId>
       <version>1.3</version>
     </dependency>
+    <dependency>
```
zheins: Remove this dependency

annotator/src/main/java/org/cbioportal/models/MutationRecord.java
```diff
@@ -507,4 +509,61 @@ public void initHeader() {
...
+            if (check.equalsIgnoreCase(ignore)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+
+    @Override
+    public boolean equals(Object o) {
```
zheins: Note to self: Only comparing fields that define a mutation here - not every field is checked.

annotator/src/main/java/org/cbioportal/models/MutationRecord.java
```diff
@@ -507,4 +509,61 @@ public void initHeader() {
         header.add("n_ref_count");
         header.add("n_alt_count");
     }
+
+    private static List<String> initMissingValueList() {
+        List<String> missingValueList = new ArrayList<>();
+        missingValueList.add("NA");
+        missingValueList.add("N/A");
+        missingValueList.add("N/a");
+        missingValueList.add("n/A");
```
zheins: I think it would be better to have an enum or something, or to do this check in a case insensitive way.

sheridancbio:
Closed : these changes have already been incorporated into the genome-nexus server

PR #41, opened 2017-08-23
Updated properties and url construction in annotation pipeline

angelicaochoa:

### DIFF
--- a/annotationPipeline/src/main/resources/application.properties.EXAMPLE
+++ b/annotationPipeline/src/main/resources/application.properties.EXAMPLE
@@ -2,5 +2,5 @@ spring.batch.job.enabled=false
 genomenexus.hgvs=<HOST>/variant_annotation/hgvs/
 genomenexus.xrefs=<HOST>/xrefs/
 genomenexus.isoform_query_parameter=isoformOverrideSource
-genomenexus.hotspot_parameter=summary
+genomenexus.enrichment_fields=fields=hotspots,mutation_assessor
 chunk=10
--- a/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
+++ b/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
@@ -70,8 +70,8 @@ public class GenomeNexusImpl implements Annotator {
     private String hgvsServiceUrl;
     @Value("${genomenexus.isoform_query_parameter}")
     private String isoformQueryParameter;
-    @Value("${genomenexus.hotspot_parameter}")
-    private String hotspotParameter;
+    @Value("${genomenexus.enrichment_fields}")
+    private String enrichmentFields;
     @Value("${genomenexus.xrefs}")
     private String geneXrefsServiceUrl;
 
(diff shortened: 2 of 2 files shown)


annotationPipeline/src/main/resources/application.properties.EXAMPLE
```diff
@@ -2,5 +2,5 @@ spring.batch.job.enabled=false
 genomenexus.hgvs=<HOST>/variant_annotation/hgvs/
 genomenexus.xrefs=<HOST>/xrefs/
 genomenexus.isoform_query_parameter=isoformOverrideSource
-genomenexus.hotspot_parameter=summary
+genomenexus.hotspot_mutassessor_parameter=fields=hotspots,mutation_assessor
```
inodb: this is enrichments parameter right? So maybe call it something like `genomenexus.enrichments_parameters`. Or if you want to specifically just set fields, maybe `genomenexus.enrichments_fields`. And then always include `fields=` in the parameter

inodb:
@angelicaochoa I think it would be good to have some test for this. If you send me a MAF that you usually use for testing (ideally public data) I could give it a shot. It's prolly out of the scope of this PR to add an automated test, but I'd anyway would like to be able to manually check. Ofc don't let that block the merging of the PR if this is urgent

angelicaochoa:
@inodb Can we have a follow up PR for testing? I think Ritika is actively using the annotation pipeline rn. I pointed her to jar for her to use in the meantime with these changes

inodb:
@angelicaochoa ok :+1:

annotationPipeline/src/main/resources/application.properties.EXAMPLE
angelicaochoa: 👍 much better, I'll update in a moment

annotationPipeline/src/main/resources/application.properties.EXAMPLE
```diff
@@ -2,5 +2,5 @@ spring.batch.job.enabled=false
 genomenexus.hgvs=<HOST>/variant_annotation/hgvs/
 genomenexus.xrefs=<HOST>/xrefs/
 genomenexus.isoform_query_parameter=isoformOverrideSource
-genomenexus.hotspot_parameter=summary
+genomenexus.enrichment_fields=fields=hotspots,mutation_assessor
```
sheridancbio: the double equals sign looks risky, but I guess it is ok if the properties file parser breaks the key/value at the first equals sign. so the URL below is going to look like .............&fields=hotspots,mutation_assessor
which makes more sense than .........&summary=summary

inodb: @angelicaochoa Sorry, this isn't entirely what I meant. Given that the prop is called `enrichment_fields`, it seems redundant to include `fields=` in the value. Better to update the code from `"&" + enrichmentFields` to `"&fields=" + enrichmentFields`. Not that important tho

sheridancbio: I've made #42 to do the change suggested by @inodb

inodb: Thanks!

PR #42, opened 2017-08-24
simplify the enrichment_fields parameter

sheridancbio:
- remove the url parameter "fields=" from the property


sheridancbio:
If this is merged, the properties for genome-nexus-annotation-pipeline must be updated to match the change to application.properties.EXAMPLE

PR #43, opened 2017-08-24
Updated database annotator pipeline

angelicaochoa:
* Removed unused oracle dependency from pom.
* Fixed casing issue in the processor that prevented the start and end genomic coordinates from being set properly when being converted to an`AnnotatedRecord` object.
* Updated example `application.properties`

### DIFF
--- a/databaseAnnotator/src/main/java/org/cbioportal/database/annotator/AnnotateRecordsProcessor.java
+++ b/databaseAnnotator/src/main/java/org/cbioportal/database/annotator/AnnotateRecordsProcessor.java
@@ -80,8 +80,8 @@ public class AnnotateRecordsProcessor implements ItemProcessor<MutationEvent, Mu
             chromosome = i.getCHR();
         }
         mafLine.put("Chromosome", chromosome);
-        mafLine.put("START_POSITION", String.valueOf(i.getSTART_POSITION()));
-        mafLine.put("END_POSITION", String.valueOf(i.getEND_POSITION()));
+        mafLine.put("Start_Position", String.valueOf(i.getSTART_POSITION()));
+        mafLine.put("End_Position", String.valueOf(i.getEND_POSITION()));
         mafLine.put("Reference_Allele", i.getREFERENCE_ALLELE());
         mafLine.put("Tumor_Seq_Allele2", i.getTUMOR_SEQ_ALLELE());
         MutationRecord record = annotator.createRecord(mafLine);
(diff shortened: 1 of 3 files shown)


databaseAnnotator/src/main/java/org/cbioportal/database/annotator/AnnotateRecordsProcessor.java
```diff
@@ -114,13 +114,6 @@ else if (annotatedRecord.getVARIANT_CLASSIFICATION().equals("5'Flank") ||
         String chr = !(annotatedRecord.getCHROMOSOME() == null || annotatedRecord.getCHROMOSOME().isEmpty()) ? annotatedRecord.getCHROMOSOME() : i.getCHR();
         Integer start = i.getSTART_POSITION();
         Integer end = i.getEND_POSITION();
-        try {
```
sheridancbio: Why is this check removed? Too much logging maybe? Are there many non-integer start/end positions? (NA's)? I would think that this must be a required field.

databaseAnnotator/src/main/resources/application.properties.EXAMPLE
```diff
@@ -5,6 +5,7 @@ databaseannotator.pass=
 databaseannotator.driver=com.mysql.jdbc.Driver
 databaseannotator.connection_string=jdbc:mysql://<HOST>/<DATABASE>
 
-genomenexus.hgvs=http://<HOST>:<PORT>/variant_annotation/hgvs/
+genomenexus.hgvs=<HOST>/variant_annotation/hgvs/
+genomenexus.xrefs=<HOST>/xrefs/
 genomenexus.isoform_query_parameter=isoformOverrideSource
-genomenexus.hotspot_parameter=cancerHotspots
+genomenexus.enrichment_fields=hotspots,mutation_assessor
```
sheridancbio: This change was present in #42 .. and must be coordinated with the change to GenomeNexusImpl.java in order to work correctly, so I think this should be removed from here. [or GenomeNexusImpl changes should be added here and #42 can be dropped]

sheridancbio: @42 is now merged, so during rebase, the change to genomenexus.enrichment_fields should already be present

databaseAnnotator/src/main/java/org/cbioportal/database/annotator/AnnotateRecordsProcessor.java
angelicaochoa: Mutation events should never have null or NA values but maybe this should be added back in case we accidentally get a fusion event when querying for mutation events with `protein_change=MUTATED`.

databaseAnnotator/src/main/resources/application.properties.EXAMPLE
sheridancbio: Never mind .. I see that there are 2 properties files involved #42 changed annotationPipeline/src/main/resources/application.properties.EXAMPLE
this changes
databaseAnnotator/src/main/resources/application.properties.EXAMPLE

There are also 2 application-test.properties.EXAMPLE files which  might need to be updated.

PR #45, opened 2017-08-25
Update application.properties.EXAMPLE

jjgao:


annotationPipeline/src/main/resources/application.properties.EXAMPLE
```diff
@@ -1,5 +1,5 @@
 spring.batch.job.enabled=false
-genomenexus.hgvs=<HOST>/variant_annotation/hgvs/
+genomenexus.hgvs=<HOST>/hgvs/
```
jjgao: Should we have genomenexus.host instead?

PR #46, opened 2017-09-26
Resolving HGVSp short bug fix, optimization

angelicaochoa:
Fixed a bug when resolving HGVSp_Short to prevent IndexOutOfBound exceptions when resolving HGVSp_Short from `amino_acid_parts` field.

Optimized pipeline when option enabled for replacing hugo gene symbol and entrez gene id. Annotator now uses a hashmap to keep track of which Ensembl gene accessions we've seen to get their corresponding entrez gene ids. Annotator will now only hit Ensembl API for new gene accessions we come across during annotation.

### DIFF
--- a/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
+++ b/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
@@ -80,6 +80,7 @@ public class GenomeNexusImpl implements Annotator {
     private TranscriptConsequence canonicalTranscript;
 
     private List<GenomeNexusIsoformOverridesResponse> overrides = new ArrayList<>();
+    private Map<String, String> ensemblAccessionEntrezIdMap = new HashMap<>();
 
     private final Logger log = Logger.getLogger(GenomeNexusImpl.class);
 
@@ -385,8 +386,17 @@ public class GenomeNexusImpl implements Annotator {
         try {
             String[] aaParts = canonicalTranscript.getAminoAcids().split("/");
             if (canonicalTranscript.getConsequenceTerms() != null && canonicalTranscript.getConsequenceTerms().get(0).equals("inframe_insertion")) {
-                hgvsp = aaParts[1].substring(0,1) + canonicalTranscript.getProteinStart() + "_" + aaParts[1].substring(1, 2) + "ins" +
-                        canonicalTranscript.getProteinEnd() + aaParts[1].substring(2);
+                // to prevent IndexOutOfBoundsException's we check for 'dup' in the HGVSc field (ex: ENST00000357654.3:c.5266dupC)
+                // since the 'amino_acids' may not always be provided in such a way where the reference allele is available
+                // ex: with the reference aa 'N/KN' and without the reference aa '-/K'. 
+                // the second format will throw an IndexOutOfBoundsException when trying to access substring index 1 
+                if (canonicalTranscript.getHgvsc() != null && canonicalTranscript.getHgvsc().contains("dup")) {
+                    hgvsp = aaParts[1].substring(0,1) + String.valueOf(Integer.valueOf(canonicalTranscript.getProteinStart()) - 1 ) + "dup";
+                }
+                else {
+                    hgvsp = aaParts[1].substring(0,1) + canonicalTranscript.getProteinStart() + "_" + aaParts[1].substring(1, 2) + "ins" +
+                            canonicalTranscript.getProteinEnd() + aaParts[1].substring(2);
+                }
             }
             else if (canonicalTranscript.getConsequenceTerms() != null && canonicalTranscript.getConsequenceTerms().get(0).equals("inframe_deletion")) {
                 hgvsp = aaParts[0] + "del";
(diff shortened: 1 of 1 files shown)


annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -385,8 +386,13 @@ private String resolveHgvspShortFromAAs() {
         try {
             String[] aaParts = canonicalTranscript.getAminoAcids().split("/");
             if (canonicalTranscript.getConsequenceTerms() != null && canonicalTranscript.getConsequenceTerms().get(0).equals("inframe_insertion")) {
-                hgvsp = aaParts[1].substring(0,1) + canonicalTranscript.getProteinStart() + "_" + aaParts[1].substring(1, 2) + "ins" +
-                        canonicalTranscript.getProteinEnd() + aaParts[1].substring(2);
+                if (canonicalTranscript.getHgvsc() != null && canonicalTranscript.getHgvsc().contains("dup")) {
+                    hgvsp = aaParts[1].substring(0,1) + String.valueOf(Integer.valueOf(canonicalTranscript.getProteinStart()) - 1 ) + "dup";
```
sheridancbio: I'm sure this is a larger issue (outside the scope of this PR) but I think that having string parsing of attributes in client code is a sign of fragility. If we need to be breaking up the getAminoAcids() result into subfields and taking substrings of those fields we probably should implement this logic in a model class or in a parser class. For one thing it would make reading / comprehending this code easier, since it is not obvious what is in the first character of the second field .. so a named method would be self-documenting.

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -515,19 +521,40 @@ private String resolveEntrezGeneId(boolean replace) {
         if (!replace || canonicalTranscript == null || canonicalTranscript.getGeneId() == null || canonicalTranscript.getGeneId().trim().isEmpty()) {
             return mRecord.getENTREZ_GENE_ID();
         }
+        // check hashmap first before hitting api for gene cross-ref data
+        if (ensemblAccessionEntrezIdMap.containsKey(canonicalTranscript.getGeneId())) {
```
sheridancbio: I guess it is true that every ensemblAccession maps to only a single EntrezId?

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -515,19 +521,40 @@ private String resolveEntrezGeneId(boolean replace) {
...
-                break;
+            if (xref.getDbname().equals("EntrezGene") && xref.getDisplayId().equals(canonicalTranscript.getGeneSymbol())) {
+                geneXrefs.add(xref);
             }
         }
-        return entrezGeneId != null ? entrezGeneId : mRecord.getENTREZ_GENE_ID();
+        if (geneXrefs.size() > 1 ) {
+            for (GeneXref xref : geneXrefs) {
+                if (!xref.getDescription().toLowerCase().contains("pseudogene") && !xref.getDescription().toLowerCase().contains("uncharacterized")) {
+                    entrezGeneId = xref.getPrimaryId();
```
sheridancbio: Here it seems that if multiple entrez genes [ignoring pseudo genes and uncharacterized] map to the ensembl id, we are going to be taking the first one encountered in the body, and not logging the fact that others were ignored. Maybe this cannot happen?

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -385,8 +386,13 @@ private String resolveHgvspShortFromAAs() {
         try {
             String[] aaParts = canonicalTranscript.getAminoAcids().split("/");
             if (canonicalTranscript.getConsequenceTerms() != null && canonicalTranscript.getConsequenceTerms().get(0).equals("inframe_insertion")) {
-                hgvsp = aaParts[1].substring(0,1) + canonicalTranscript.getProteinStart() + "_" + aaParts[1].substring(1, 2) + "ins" +
-                        canonicalTranscript.getProteinEnd() + aaParts[1].substring(2);
+                if (canonicalTranscript.getHgvsc() != null && canonicalTranscript.getHgvsc().contains("dup")) {
```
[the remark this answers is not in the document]
angelicaochoa: An example of this is `ENST00000357654.3:c.5266dupC` . "dup" will be followed by the duplicated nucleotide(s)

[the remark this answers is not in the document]
angelicaochoa: (also only occurring for in frame insertions)

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
angelicaochoa: yes, as long as it's not a pseudogene or uncharacterized area of the genome then it should be one entrez gene id per ensembl gene accession

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
n1zea144: I think having some examples of Hgvsc and amino acid parts in comments would be helpful.

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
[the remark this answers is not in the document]
angelicaochoa: in any case, "dup" should not appear by coincidence either

angelicaochoa:
@n1zea144 I added some comments to document the check for `dup` in the returned HGVSc value

PR #48, opened 2017-10-11
HGVSc bug fix.

angelicaochoa:


sheridancbio:
This is simpler and fixes a bug in the Ceil() approach. The previous code was resulting in ppos of 2 when cpos was 2 ; also ppos of 1 when cpos was 3.
New code give ppos of 1 when cpos is in  {1,2,3} ; and ppos of 2 when cpos is in {4,5,6}, etc.

PR #49, opened 2017-10-31
Log progress of annotation

mandawilson:


annotationPipeline/src/main/java/org/cbioportal/annotation/pipeline/MutationRecordReader.java
```diff
@@ -116,7 +116,14 @@ public void handleLine(String line) {
         }
         reader.close();
 
-        for(MutationRecord record : mutationRecords) {
+        int variantsToAnnotateCount = mutationRecords.size();
+        int annotatedVariantsCount = 0;
+        LOG.info(String.valueOf(variantsToAnnotateCount) + " records to annotate");
+        for (MutationRecord record : mutationRecords) {
+            annotatedVariantsCount++;
+            if (annotatedVariantsCount % 2000 == 0) {
```
sheridancbio: When annotations are not in the GN cache, they can take several seconds each. If they take 3 seconds per annotation that would mean 6000 seconds between progress updates (almost 2 hours). I would recommend a shorter reporting interval (400 ?).
Maybe this will lead to too many log messages when annotations are done fast from cache ... but the rate could be limited by adding something like:
<pre>
public static final int VARIANTS_BETWEEN_PROGRESS_REPORTS = 100;
public static final long MILLISECONDS_BETWEEN_PROGRESS_REPORTS = 1000 * 60 * 10; // 10 minutes
...
Instant nextProgressReportTime = Instant.now().plusMillis(MILLISECODS_BETWEEN_REPORTS);
...
if (annotatedVariantsCount % VARIANTS_BETWEEN_PROGRESS_REPORTS == 0 && Instant.now().isAfter(nextProgressReportTime)) {
    nextProgressReportTime = nextProgressReportTime.plusMillis(MILLISECODS_BETWEEN_REPORTS);
    LOG.info(...)
...
</pre>

PR #52, opened 2017-12-13
Add circle CI test for checking if MAF is in sync

inodb:
Enable Circle CI (similar to Travis CI, but faster).

This adds a test:
- annotate first 100 records of mskimpact2017 using mskimpact transcripts
- annotate first 100 records of mskimpact2017 using uniprot transcripts

See example: https://circleci.com/gh/genome-nexus/genome-nexus-annotation-pipeline/4


inodb:
@sheridancbio i remember there were some particular variants you'd like to see included in the test?

sheridancbio:
sorry @inodb .. I cannot remember a list of particular mutations that should be run through the annotation pipeline, so don't let me hold things up.

inodb:
@sheridancbio good points regarding take all kind of classes/types of mutations and all chromosomes. Maybe also testing e.g. all oncokb annotated mutations would be good. We can tackle that in some follow up PR

It is indeed an integration test, that checks whether annotation works with current production genome nexus API. It's p fast on 100 mutations (only takes about 20s as you can see in the circle-ci test log:  https://circleci.com/gh/genome-nexus/genome-nexus-annotation-pipeline/4)

I'm not sure about the logic of the annotator. Is re-annotation different from annotating an unannotated MAF file?

PR #55, opened 2018-01-10
Bug fix for IF indels miscategorized as FS indels.

angelicaochoa:
A parenthesis was misplaced when handling frame shift events, which helped miscategorize In Frame indels as Frame Shift indels. 

The way that a variant is determined to be in frame or not was also modified to ensure that we correctly identify in frame events.

### DIFF
--- a/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
+++ b/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
@@ -719,10 +719,12 @@ public class GenomeNexusImpl implements Annotator {
     }
 
     private String getVariantClassificationFromMap(String variant) {
-        boolean inframe = Math.abs(mRecord.getREFERENCE_ALLELE().length() - getTumorSeqAllele(mRecord).length()) % 3 == 0;
+        String refAllele = mRecord.getREFERENCE_ALLELE().equals("-") ? "" : mRecord.getREFERENCE_ALLELE();
+        String tumorSeqAllele = getTumorSeqAllele(mRecord).equals("-") ? "" : getTumorSeqAllele(mRecord);
+        boolean inframe = Math.abs(refAllele.length() - tumorSeqAllele.length()) % 3 == 0;
         variant = variant.toLowerCase();
         String resolvedVariantType = resolveVariantType();
-        if ((variant.equals("frameshift_variant") || (variant.equals("protein_altering_variant") || variant.equals("coding_sequence_variant")) && !inframe)) {
+        if ((variant.equals("frameshift_variant") || variant.equals("protein_altering_variant") || variant.equals("coding_sequence_variant")) && !inframe) {
             if (resolvedVariantType.equals("DEL")) {
                 return "Frame_Shift_Del";
             }


annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -719,10 +719,12 @@ private TranscriptConsequence transcriptWithMostSevereConsequence(List<Transcrip
     }
 
     private String getVariantClassificationFromMap(String variant) {
-        boolean inframe = Math.abs(mRecord.getREFERENCE_ALLELE().length() - getTumorSeqAllele(mRecord).length()) % 3 == 0;
+        String refAllele = mRecord.getREFERENCE_ALLELE().equals("-") ? "" : mRecord.getREFERENCE_ALLELE();
```
sheridancbio: So on an insert the reference_allele field is exactly one hyphen ; and on a delete the tumor_seq_allele is exactly one hyphen. This extra hyphen (when present) must be stripped before doing the modulo 3 math.

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -719,10 +719,12 @@ private TranscriptConsequence transcriptWithMostSevereConsequence(List<Transcrip
...
+        String tumorSeqAllele = getTumorSeqAllele(mRecord).equals("-") ? "" : getTumorSeqAllele(mRecord);
+        boolean inframe = Math.abs(refAllele.length() - tumorSeqAllele.length()) % 3 == 0;
         variant = variant.toLowerCase();
         String resolvedVariantType = resolveVariantType();
-        if ((variant.equals("frameshift_variant") || (variant.equals("protein_altering_variant") || variant.equals("coding_sequence_variant")) && !inframe)) {
+        if ((variant.equals("frameshift_variant") || variant.equals("protein_altering_variant") || variant.equals("coding_sequence_variant")) && !inframe) {
```
sheridancbio: I'm not sure of the meaning of "protein_altering_variant" ... would a deletion of three (coding) nucleotides be called a "protein_altering_variant" by GN or a "coding_sequence_variant" ? My guess: "protein_altering_variant".

I'm also confused about what circumstances would cause GN to call something "frameshift_variant" but then we overrule GN with our own modulo three mathematics.  Maybe we know that GN sometimes gets this wrong in the annotation even though it gets the NT changes correct?

angelicaochoa: @sheridancbio Genome Nexus doesn't determine which ENSEMBL consequence term is returned - that's returned by the VEP service. There are also priorities given to consequence terms so if a variant can be categorized as a few different consequence terms, then I believe VEP assigns the consequence term with the (highest priority)[http://useast.ensembl.org/info/genome/variation/predicted_data.html#consequences] to the variant. 

What GN does is it maps ENSEMBL consequence terms to the cBioPortal equivalent, and takes into consideration (1) the difference in length between the ref allele and alt alleles to determine if the event is in frame or not and (2) the resolved variant type (insertion, deletion, snp, etc.)

PR #56, opened 2018-01-11
Unit tests for resolved mutation types, protein changes.

angelicaochoa:


annotator/src/test/java/org/cbioportal/annotator/internal/GenomeNexusImplTest.java
```diff
@@ -0,0 +1,331 @@
...
+    }
+
+    private List<AnnotatedRecord> makeMockAnnotatedRecords() {
+        List<MutationRecord> mockMutationRecords = makeMockMutationRecords();
+
+        List<AnnotatedRecord> mockAnnotatedRecords = new ArrayList();
+        for (MutationRecord record : mockMutationRecords) {
+            AnnotatedRecord annotatedRecord = null;
+            try {
+                annotatedRecord = annotator.annotateRecord(record, true, ISOFORM_OVERRIDE_SOURCE, true);
```
sheridancbio: This seems like an actual call to Genome Nexus to annotate records -- for a unit test, I generally think that we want to isolate the tested code from any dependencies. Whether Genome Nexus is available or not, or a change to the genome nexus annotation behavior probably should not cause unit tests to fail here. I'm not as familiar with how the annotator codebase is organized, but I think I would recommend either (1) encapsulating the functionality which sends requests to the genome nexus web service into its own class (a @Repository?) and then mocking that class to return pre-computed annotated records to support this unit test, or (2) isolating the business logic that transforms the genome nexus annotation into AnnotatedRecords (a @Service?) and then testing only that business logic by passing in the annotations coming from Genome Nexus.
We can also create integration tests which confirm the proper interactions occur between the genome nexus web server and the genome nexus annotation client ... these would be valuable. But the focus of the integration tests should be to hit every interface dependency (all endpoints) using all valid argument combinations. And the integration tests should not be automatically run during every code build or PR submission, but should be run periodically (probably manually with 'mvn integration-test' or similar)
That's the picture in my mind ... but happy to discuss options.

pom.xml
```diff
@@ -31,16 +31,14 @@
...
-		<groupId>org.springframework.boot</groupId>
-		<artifactId>spring-boot-starter-test</artifactId>
-	</dependency>
-
-	<dependency>
-		<groupId>org.springframework.batch</groupId>
-		<artifactId>spring-batch-test</artifactId>
-	</dependency>
+    <dependency>
+        <groupId>org.springframework.boot</groupId>
```
sheridancbio: Thank you for getting rid of the tabs here. Maybe standardize to 2 space indenting as well?

annotator/src/test/java/org/cbioportal/annotator/internal/GenomeNexusImplTest.java
angelicaochoa: thanks @sheridancbio I think that's a good idea to mock the GN responses

mandawilson:
This is wonderful.  Optionally, maybe use Mockito so that we can use the real GenomeNexusImpl and just mock a few methods of it (in this case I think just one method).  Instead of MockGenomeNexusImpl.java maybe make a new method with GenomeNexusImpl.java lines 131 to 136 (`new RestTemplate()`  through `gnResponseBody = responseEntity.getBody()`) and then just mock that new method.  I would use `GenomeNexus genomeNexus = Mockito.mock(GenomeNexusImpl.class)` and then `genomeNexus.when(NEW_METHOD(urlForRecord)).thenReturn(CORRECT_GenomeNexusAnnotationResponse_ARRAY)`.  I would be happy to do it as a PR on your PR if you want, just for fun.  But only if we all agree that is better and we have the spare time.  I think this is a really great start, so we could just make a note that maybe we could think about doing it later.

PR #57, opened 2018-02-09
Catch additional types of splice site variants.

angelicaochoa:

### DIFF
--- a/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
+++ b/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
@@ -379,7 +379,7 @@ public class GenomeNexusImpl implements Annotator {
                     }
                 }
             }
-            else if (canonicalTranscript.getHgvsc() != null && spliceSiteVariants.contains(canonicalTranscript.getConsequenceTerms().get(0))) {
+            else if (canonicalTranscript.getHgvsc() != null) {
                 Integer cPos = 0;
                 Integer pPos = 0;
                 Matcher m = cDnaExtractor.matcher(canonicalTranscript.getHgvsc());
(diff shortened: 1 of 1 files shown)


annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -91,7 +91,7 @@
         {"Thr", "T"}, {"Trp", "W"}, {"Tyr", "Y"}, {"Val", "V"}, {"Xxx", "X"}, {"Ter", "*"}
     };
     private Set<String> spliceSiteVariants = new HashSet<String>(Arrays.asList(
-            "splice_acceptor_variant", "splice_donor_variant", "splice_region_variant"));
+            "splice_acceptor_variant", "splice_donor_variant", "splice_region_variant", "stop_lost"));
```
ckandoth: `stop_lost` is not a type of spliceSiteVariant. Better to handle it separately or not at all, as described below.

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -379,7 +379,10 @@ private String resolveHgvspShort() {
                     }
                 }
             }
-            else if (canonicalTranscript.getHgvsc() != null && spliceSiteVariants.contains(canonicalTranscript.getConsequenceTerms().get(0))) {
+            else if (canonicalTranscript.getHgvsc() != null && (spliceSiteVariants.contains(canonicalTranscript.getConsequenceTerms().get(0)) ||
```
ckandoth: It is safe to do this if `canonicalTranscript.getHgvsc() != null` and nothing else.

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -369,6 +369,7 @@ private String resolveHgvsp() {
     }
 
     private String resolveHgvspShort() {
+        String variantClassification = resolveVariantClassification();
```
n1zea144: You don't use this until the nested else if (line 396) - I suggest you move it down to where it is used.

PR #58, opened 2018-02-23
Test vcf2maf test.maf file

inodb:
Create a test that uses vcf2maf test.maf file and checks whether the output of
genome nexus is the same as that from vcf2maf.

Only compares the columns 
```
$ csvcut -tn test/data/vcf2maf/tests.maf.gn_output.uniprot.cut_columns.txt
  1: Hugo_Symbol
  2: Chromosome
  3: Start_Position
  4: End_Position
  5: Strand
  6: Variant_Classification
  7: Variant_Type
  8: Reference_Allele
  9: Tumor_Seq_Allele1
 10: Tumor_Seq_Allele2
 11: Tumor_Sample_Barcode
 12: Matched_Norm_Sample_Barcode
 13: Match_Norm_Seq_Allele1
 14: Match_Norm_Seq_Allele2
 15: Tumor_Validation_Allele1
 16: Tumor_Validation_Allele2
 17: Match_Norm_Validation_Allele1
 18: Match_Norm_Validation_Allele2
 19: Verification_Status
 20: Validation_Status
 21: Mutation_Status
 22: Sequencing_Phase
 23: Sequence_Source
 24: Validation_Method
 25: Score
 26: BAM_File
 27: Sequencer
 28: t_ref_count
 29: t_alt_count
 30: n_ref_count
 31: n_alt_count
 32: n_depth
 33: t_depth
```

PR #59, opened 2018-02-23
Check for ambiguous (SNP/INDEL) variant allele.

angelicaochoa:
# What? Why?
Due to bugs in data, it is possible for a MAF record to contain variant alleles: `Tumor_Seq_Allele1`="-" and `Tumor_Seq_Allele2`=<some snp>  (or vice versa).

In these cases it is ambiguous which tumor allele to choose since neither are equal to `Reference_Allele`, which is how it is currently resolved with the chosen variant allele being the one which is not equal to `Reference_Allele`. 

Changes proposed in this pull request:
- Resolve ambiguous variant allele by giving priority to SNPs over "-"

Other changes:
- Standardized logging and added `log4j.properties.EXAMPLE`
- Added cbioportal/core dependency
- Added verbose argument to annotation pipeline to suppress annotation failures by default (made logs too cluttered)
- Changed version to `1.0.0`

cbioportal/cbioportal pr: https://github.com/cBioPortal/cbioportal/pull/3913


pieterlukasse:
would be great if https://github.com/genome-nexus/genome-nexus-annotation-pipeline/pull/58 can be merged and this one rebased on top of that. This will trigger automatic tests.

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -596,11 +600,11 @@ public String convertToHgvs(MutationRecord record)
         String start = record.getSTART_POSITION();
         String end = record.getEND_POSITION();
         String ref = record.getREFERENCE_ALLELE();
-        String var = getTumorSeqAllele(record);
+        String var = MafUtil.resolveTumorSeqAllele(record.getREFERENCE_ALLELE(), record.getTUMOR_SEQ_ALLELE1(), record.getTUMOR_SEQ_ALLELE2());
 
         String prefix = "";
         if(ref.equals(var)) {
-            log.warn("Warning: Reference allele extracted from " + chr + ":" + start + "-" + end + " matches alt allele. Sample: " + record.getTUMOR_SAMPLE_BARCODE());
+            LOG.warn("Warning: Reference allele extracted from " + chr + ":" + start + "-" + end + " matches alt allele. Sample: " + record.getTUMOR_SAMPLE_BARCODE());
```
pieterlukasse: shouldn't this be logged as an ERROR instead of WARNING? Should these cases be filtered out when loading in cBioPortal?

annotationPipeline/pom.xml
```diff
@@ -4,12 +4,12 @@
   <name>cBioPortal Annotation Pipeline</name>
   <description>Annotation Pipeline</description>
   <artifactId>annotationPipeline</artifactId>
-  <version>0.1.0</version>
+  <version>1.0.0</version>
```
sheridancbio: I guess we have not been versioning the code yet, so this is good. But why a major revision increase? I guess it doesn't matter, but there is no fundamental change to the code, so we could use 0.2.0

annotationPipeline/src/main/java/org/cbioportal/annotation/pipeline/MutationRecordReader.java
```diff
@@ -131,61 +137,63 @@ public void handleLine(String line) {
...
+            header.addAll(annotatedRecord.getHeaderWithAdditionalFields());
+            
+            // log server failure message if applicable
+            if (!serverErrorMessage.isEmpty()) {
+                LOG.warn(serverErrorMessage);
+                if (errorReportLocation != null) updateErrorMessages(record, record.getVARIANT_CLASSIFICATION(), annotator.getUrlForRecord(record, isoformOverride), serverErrorMessage);
+                continue;
+            }
+            String annotationErrorMessage = "";
+            if (MafUtil.variantContainsSnpAndIndel(record.getREFERENCE_ALLELE(), record.getTUMOR_SEQ_ALLELE1(), record.getTUMOR_SEQ_ALLELE2())) {
```
sheridancbio: As the cbioportal core function is currently implemented, I'm not sure if the use here covers all desired cases. The core function looks for tumorSeqAllele1.equals("-") ... but theoretically a case like this would not be detected for error logging:
refAllele: "AAACCCGGG" (codes for "KPG")
tumorSeqAllele1 : "AAA-GGG" (codes for "KG")
tumorSeqAllele2 : "AAATCCGGG" (codes for "KSG")
Maybe this is very unlikely to appear in the input? Also I'm not sure which mutation should be selected and which should be ignored (The SNP, or the in-frame deletion)
Or maybe this will never appear in the input .. because any flanking nucleotides common to both refAllele and tumorSeqAllele1/2 will be stripped off in the reported variation ... so for the above case we would only ever get:
"CCC"
"-"
"TCC"
?

annotationPipeline/src/main/java/org/cbioportal/annotation/pipeline/MutationRecordReader.java
```diff
@@ -131,61 +137,63 @@ public void handleLine(String line) {
...
-        saveErrorMessagesToFile(errorMessages);
-
-        List<String> full_header = new ArrayList(header);
-        ec.put("mutation_header", full_header);
+        // print summary statistics and save error messages to file if applicable
+        printSummaryStatistics(failedAnnotations, failedNullHgvspAnnotations, failedMitochondrialAnnotations, snpAndIndelVariants);
+        if (errorReportLocation != null) {
+            saveErrorMessagesToFile(errorMessages);
+        }
+        ec.put("mutation_header", new ArrayList(header));
```
sheridancbio: good code reduction

annotationPipeline/src/main/java/org/cbioportal/annotation/pipeline/MutationRecordReader.java
```diff
@@ -131,61 +137,63 @@ public void handleLine(String line) {
...
-        errorMessages.add("Total records that failed annotation: " + failedAnnotations);
-        errorMessages.add("# records with HGVSp null variant classification: " + failedNullHgvspAnnotations);
-        errorMessages.add("# records with Mitochondrial variants: " + failedMitochondrialAnnotations);
-        saveErrorMessagesToFile(errorMessages);
-
-        List<String> full_header = new ArrayList(header);
-        ec.put("mutation_header", full_header);
+        // print summary statistics and save error messages to file if applicable
+        printSummaryStatistics(failedAnnotations, failedNullHgvspAnnotations, failedMitochondrialAnnotations, snpAndIndelVariants);
+        if (errorReportLocation != null) {
```
sheridancbio: this was not checked before?  Good addition.

annotationPipeline/src/main/java/org/cbioportal/annotation/pipeline/MutationRecordReader.java
```diff
@@ -202,15 +210,31 @@ public AnnotatedRecord read() throws Exception {
...
         List<String> msg = Arrays.asList(new String[]{record.getTUMOR_SAMPLE_BARCODE(), record.getCHROMOSOME(),
                                 record.getSTART_POSITION(), record.getEND_POSITION(), record.getREFERENCE_ALLELE(),
-                                record.getTUMOR_SEQ_ALLELE2(), record.getVARIANT_CLASSIFICATION(), reasonFailed, url});
+                                record.getTUMOR_SEQ_ALLELE1(), record.getTUMOR_SEQ_ALLELE2(), variantClassification, 
+                                errorMessage, url});
         errorMessages.add(StringUtils.join(msg, "\t"));
     }
 
     private void saveErrorMessagesToFile(List<String> errorMessages) {
-        if (errorReportLocation == null) return;
```
sheridancbio: aha .. this is where the check was before.

annotator/pom.xml
```diff
@@ -4,13 +4,13 @@
   <name>cBioPortal Annotator</name>
   <description>Annotator</description>
   <artifactId>annotator</artifactId>
-  <version>0.1.0</version>
+  <version>1.0.0</version>
```
sheridancbio: could be 0.2.0  (or 1.0.0 is fine too)

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -92,6 +94,8 @@
     };
     private Set<String> spliceSiteVariants = new HashSet<String>(Arrays.asList(
             "splice_acceptor_variant", "splice_donor_variant", "splice_region_variant"));
+    private Set<String> nucleotides = new HashSet<String>(Arrays.asList(
```
sheridancbio: This does not seem to be used

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -744,7 +748,8 @@ private TranscriptConsequence transcriptWithMostSevereConsequence(List<Transcrip
 
     private String getVariantClassificationFromMap(String variant) {
         String refAllele = mRecord.getREFERENCE_ALLELE().equals("-") ? "" : mRecord.getREFERENCE_ALLELE();
-        String tumorSeqAllele = getTumorSeqAllele(mRecord).equals("-") ? "" : getTumorSeqAllele(mRecord);
+        String tumorSeqAllele = MafUtil.resolveTumorSeqAllele(mRecord.getREFERENCE_ALLELE(), mRecord.getTUMOR_SEQ_ALLELE1(), mRecord.getTUMOR_SEQ_ALLELE2());
+        if (tumorSeqAllele.equals("-")) tumorSeqAllele = "";
```
sheridancbio: My interpretation of this is: we are dealing with a deletion and we are standardizing the representation to remove the placeholder character for a gap ("-"). My comment above regarding the possible problem with  .equals("-")  applies here ... unless all common flanking nucleotides are always stripped.

pom.xml
```diff
@@ -24,23 +24,57 @@
...
+    <slf4j.version>1.6.6</slf4j.version>
   </properties>
 
   <dependencies>
     <dependency>
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter-batch</artifactId>
+        <exclusions>
+          <exclusion>
+            <groupId>org.slf4j</groupId>
```
sheridancbio: We should discuss logging framework options .. and pick one and apply it across all code bases so we don't have to address these conflicts. (not a change request, just thinking out loud)

pom.xml
```diff
@@ -24,23 +24,57 @@
...
+      <artifactId>slf4j-log4j12</artifactId>
+      <version>${slf4j.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+      <version>${slf4j.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.github.angelicaochoa</groupId>
```
sheridancbio: shouldn't this be from a public repo? (maybe after we merge the cbioportal PR)
Also, the pipelines repo pulls in core in a slightly different way .. not a checksum version id, but a symbolic one. Should we use the same approach in both places? (maybe we will -- I have not looked at the related pipelines PR yet)

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
inodb: @pieterlukasse actually wondering how this works with the validator. Does it do any MAF file validation? Cause if it does, it should probably indicate these warnings as well.

pom.xml
inodb: @sheridancbio this should def point to public cbioportal repo once the PR is fixed

I think it would be good to have pipelines pull core repo in the same way. Unless there's a good reason for using symbolic links over hardcoded versions I'd say let's use jitpack there as well.

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
pieterlukasse: @inodb good point...I had already asked @dionnezaal to implement a few of these checks just to see how much issues we would find in the current batch of provisional TCGA files.

pieterlukasse: If we make them ERROR in the validation, then they would be flagged by the Circle CI at datahub as well.

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -166,7 +168,7 @@ public AnnotatedRecord convertResponseToAnnotatedRecord(boolean replace) {
                 resolveVariantClassification(),
                 resolveVariantType(),
                 resolveReferenceAllele(),
-                resolveReferenceAllele(),
+                mRecord.getTUMOR_SEQ_ALLELE1(),
```
angelicaochoa: After discussing with @inodb we decided to make `Tumor_Seq_Allele1` immutable. This will have no consequence on how the variant gets annotated since we are adding logic to handle ambiguous tumor seq allele cases (from `cbioportal/MafUtil`)

PR #60, opened 2018-03-13
Variant type bug fix.

angelicaochoa:

### DIFF
--- a/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
+++ b/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
@@ -529,7 +529,7 @@ public class GenomeNexusImpl implements Annotator {
                     return "";
             }
             String npType[] = {"SNP", "DNP", "TNP"};
-            return (refLength < 3 ? npType[refLength - 1] : "ONP");
+            return (refLength <= 3 ? npType[refLength - 1] : "ONP");
         } else {
             if (refLength < varLength) {
                 return "INS";


annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -529,7 +529,7 @@ private String getVariantType(String refAllele, String varAllele) {
                     return "";
             }
             String npType[] = {"SNP", "DNP", "TNP"};
-            return (refLength < 3 ? npType[refLength - 1] : "ONP");
+            return (refLength <= 3 ? npType[refLength - 1] : "ONP");
```
angelicaochoa: TNP's weren't being captured here when check for reflength is `< 3` vs. `<= 3`

sheridancbio:
@angelicaochoa The vcf2maf unit tests are coming back with diffs/failures it seems. output like this:
APC	5	112174758	112174761	+	Frame_Shift_Del	DEL	AAGA	GA	TUMOR	NORMAL	AAGA	AAGA
seems odd, where reference = AAGA and tumor_seq_allele2 = GA. Why isn't this just a deletion of AA from 112174758 to 112174759 ?  Maybe because of the missing tumor_seq_allele1?

inodb:
@sheridancbio @angelicaochoa any clue why the test fails?

ckandoth:
@sheridancbio These mutations can be skipped for testing Genome Nexus - https://github.com/mskcc/vcf2maf/blob/master/tests/test.vcf#L18-L21 - they were meant to test multiallelic variants given in VCF format. MAF format doesn't support multiallelic variants, so that's a valuable test for vcf2maf, but not for Genome Nexus.

PR #61, opened 2018-03-21
Fix bug in how HGVS string is constructed.

angelicaochoa:


annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -655,7 +655,7 @@ else if(var.equals("-") || var.length() == 0){
          Example SNP   : 2 216809708 216809709 CA T
          Example output: 2:g.216809708_216809709delCAinsT
          */
-        else if (ref.length() > 1) {
+        else if (ref.length() > 1 || var.length() > 1) {
```
angelicaochoa: variants where ref allele len is 1 but alt allele was > 1 were constructed as `chr:g.XXXXREF>ALT` rather than `chr:g.XXXXdelREFinsALT`

angelicaochoa: this check allows either ref or alt allele to be greater than one and constructs an hgvs as `chr:g.XXXXdelREFinsALT`

inodb:
Thanks @angelicaochoa - this looks great! Nice work on the testing!

Just one point re architecture. @onursumer moved most of this stuff over to the web service. There's a `/genomic` endpoint now for cancerhotspots (http://genomenexus.org/swagger-ui.html#/cancer45hotspots45controller in internal endpoints) (we still need to add it to public `annotate` endpoint). That way we didn't have to replicate this logic in the frontend when calling genome nexus annotation. I also think it makes sense for the annotator to call that endpoint directly after constructing ref/alt. To avoid duplicating this logic both in the annotator and the web service. If this has an immediate use - by all means merge, but would be good to move this over

angelicaochoa:
thanks @inodb @ckandoth !

@inodb let's merge this for now bc it is affecting some impact cases. it would be nice though to avoid duplicating this log in the future.

PR #63, opened 2018-04-24
Using the new Genome Nexus endpoint for annotation

onursumer:
- Using the new endpoint to directly resolve most of the annotation values
- Removed all the models related to Genome Nexus, relying on the generated Java API client models instead.
- Genome Nexus Java API client modules are now added as an external dependency. This PR will remain broken until we include them with jitpack.
- [x] Do Not Merge before merging https://github.com/genome-nexus/genome-nexus-java-api-client/pull/1


scripts/generateApiClientModules.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+wget http://central.maven.org/maven2/io/swagger/swagger-codegen-cli/2.3.1/swagger-codegen-cli-2.3.1.jar -O swagger-codegen-cli.jar
+java -jar swagger-codegen-cli.jar generate -i http://genomenexus.org/v2/api-docs -l java -c genomeNexusPublicApiClientConfig.json -o ../genomeNexusPublicApiClient
```
inodb: let's change this to take some parameter e.g. `$1`, could default to https://genomenexus.org/v2/api-docs

inodb: also it assumes this run from the directory this script is residing in, so prolly good to cd to the `SCRIPT_DIR` first, e.g.: https://github.com/cBioPortal/cbioportal-frontend/blob/master/scripts/generate_spec_files.sh#L2

annotationPipeline/pom.xml
```diff
@@ -12,6 +12,10 @@
     <version>1.0.0</version>
   </parent>
 
+  <properties>
+    <start-class>org.cbioportal.annotation.AnnotationPipeline</start-class>
```
onursumer: Seems like this is required to prevent maven build failure after adding the generated client files.

annotationPipeline/src/main/resources/application.properties.EXAMPLE
```diff
@@ -1,6 +1,5 @@
 spring.batch.job.enabled=false
-genomenexus.hgvs=<HOST>/hgvs/
-genomenexus.xrefs=<HOST>/xrefs/
+genomenexus.base=<HOST>
 genomenexus.isoform_query_parameter=isoformOverrideSource
-genomenexus.enrichment_fields=hotspots,mutation_assessor
+genomenexus.enrichment_fields=annotation_summary
```
onursumer: now we only rely on the annotation_summary (field name is subject to change)

annotator/src/test/java/org/cbioportal/annotator/GenomeNexusTestConfiguration.java
```diff
@@ -41,8 +41,7 @@
 @Configuration
 public class GenomeNexusTestConfiguration {
 
-    public final static String HGVS_SERVICE_URL = "http://annotation.genomenexus.org/hgvs/";
-    public final static String GENE_XREFS_SERVICE_URL = "http://annotation.genomenexus.org/xrefs/";
+    public final static String GENOME_NEXUS_BASE_URL = "http://annotation.genomenexus.org/";
```
onursumer: now we only need to pass the base URL to the API client instance

annotator/src/test/java/org/cbioportal/annotator/mixin/TranscriptConsequenceMixin.java
```diff
@@ -0,0 +1,68 @@
+
+package org.cbioportal.annotator.mixin;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.List;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class TranscriptConsequenceMixin {
```
onursumer: This mixin is only needed within the test scope. All other mappings are handled by the auto generated API client.

annotator/src/test/java/org/cbioportal/annotator/mixin/VariantAnnotationMixin.java
```diff
@@ -0,0 +1,35 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.genome_nexus.client.TranscriptConsequence;
+import org.genome_nexus.client.VariantAnnotationSummary;
+
+import java.util.List;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class VariantAnnotationMixin
```
onursumer: This mixin is only needed within the test scope. All other mappings are handled by the auto generated API client.

annotator/src/main/java/org/cbioportal/models/GenomeNexusAnnotationResponse.java
```diff
@@ -1,320 +0,0 @@
...
-    "assembly_name",
-    "seq_region_name",
-    "start",
-    "end",
-    "allele_string",
-    "strand",
-    "most_severe_consequence",
-    "transcript_consequences"
-})
-public class GenomeNexusAnnotationResponse {
```
onursumer: We don't need this model anymore, we are relying on generated API client models

annotator/src/main/java/org/cbioportal/models/GenomeNexusIsoformOverridesResponse.java
```diff
@@ -1,180 +0,0 @@
...
-
-@JsonInclude(JsonInclude.Include.NON_NULL)
-@Generated("org.jsonschema2pojo")
-@JsonPropertyOrder({
-    "transcriptId",
-    "geneSymbol",
-    "refseqId",
-    "ccdsId"
-})
-public class GenomeNexusIsoformOverridesResponse {
```
onursumer: We don't need this model either, using generated model instead

pom.xml
```diff
@@ -69,6 +70,17 @@
       <artifactId>cbioportal</artifactId>
       <version>b6b7b3f8332b94f01adc9e9263af8f2fe2b4e09a</version>
     </dependency>
+    <!-- api client -->
```
onursumer: These api client dependencies only work locally for now, we should replace this with the proper jitpack version once we merge https://github.com/genome-nexus/genome-nexus-java-api-client/pull/1

onursumer: @inodb replaced local dependencies with the jitpack ones, I think this PR is ready to merge.

annotator/pom.xml
```diff
@@ -14,6 +14,11 @@
   </parent>
 
   <dependencies>
+    <dependency>
+      <groupId>org.genome_nexus.client</groupId>
+      <artifactId>genomeNexusPublicApiClient</artifactId>
+      <version>1.0.0</version>
```
inodb: @onursumer do we need this? Seems like this has been replaced by jitpack artifacts no?

annotator/src/test/java/org/cbioportal/annotator/mixin/TranscriptConsequenceMixin.java
inodb: @onursumer can't we import this class from genome-nexus?

annotator/src/test/java/org/cbioportal/annotator/mixin/VariantAnnotationMixin.java
inodb: same here

annotator/src/test/java/org/cbioportal/annotator/mixin/TranscriptConsequenceMixin.java
onursumer: We have something similar in genome nexus, but not the same mixin. We are using this mixin to map mock data onto an instance in tests. So, not sure if it is worth importing from Genome Nexus.

annotationPipeline/pom.xml
inodb: @onursumer do we still need this now that we are importing the client files from java api client repo?

annotationPipeline/src/main/resources/application.properties.EXAMPLE
onursumer: Default value makes sense.

Although it is good to have some error handling, I wouldn't worry much about invalid enrichment fields. Genome Nexus just ignores the field if it is not valid. Also, currently we don't really need any other field than `annotation_summary`.

inodb: 👌

annotator/src/test/java/org/cbioportal/annotator/mixin/TranscriptConsequenceMixin.java
inodb: 👌

annotator/pom.xml
onursumer: @inodb good catch, this should be removed.

annotationPipeline/pom.xml
onursumer: I think we can remove this one too

onursumer: @inodb seems like we still need this for some reason, let's keep it for now.

PR #84, opened 2019-03-01
add gnomad data into file

leexgh:
Add gnoamd data into the output file:
- ColocatedVariants_gnomad_afr_maf
- ColocatedVariants_gnomad_eas_maf
- ColocatedVariants_gnomad_nfe_maf
- ColocatedVariants_dbSnpId

### DIFF
--- a/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
+++ b/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
@@ -563,8 +563,8 @@ public class GenomeNexusImpl implements Annotator {
             // get TUMOR_SEQ_ALLELE1 from mutation record
             String tumorSeqAllele1 = record.getTUMOR_SEQ_ALLELE1();
 
+            // TODO: move this to the genome nexus service (https://github.com/genome-nexus/genome-nexus/issues/254)
             // check if the dbSnpId start with "rs", now we have multiple ids returned in dbSnpId, such as "CM112509" or "COSM476".
-            // this checking can be removed after fixing the genome nexus model
             if (variant.getDbSnpId().startsWith("rs")) {
                 // check if the gnomad_nfe_allele, gnomad_afr_allele and gnomad_eas_allele matches the tumorSeqAllele1
                 if (tumorSeqAllele1 != null && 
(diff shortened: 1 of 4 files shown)


annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -507,6 +554,33 @@ private TranscriptConsequenceSummary getCanonicalTranscript(VariantAnnotation gn
...
+            String tumorSeqAllele1 = record.getTUMOR_SEQ_ALLELE1();
+
+            // check if the dbSnpId start with "rs", now we have multiple ids returned in dbSnpId, such as "CM112509" or "COSM476".
+            // this checking can be removed after fixing the genome nexus model
+            if (variant.getDbSnpId().startsWith("rs")) {
+                // check if the gnomad_nfe_allele, gnomad_afr_allele and gnomad_eas_allele matches the tumorSeqAllele1
+                if (tumorSeqAllele1 != null && 
+                    (variant.getGnomadAfrAllele() != null && !tumorSeqAllele1.equals(variant.getGnomadAfrAllele()) ||
+                    variant.getGnomadEasAllele() != null && !tumorSeqAllele1.equals(variant.getGnomadEasAllele()) ||
+                    variant.getGnomadNfeAllele() != null && !tumorSeqAllele1.equals(variant.getGnomadNfeAllele()))) {
```
inodb: @leexgh it would be good to have this logic in the web service (genome-nexus/genome-nexus) instead

pom.xml
```diff
@@ -74,12 +74,12 @@
     <dependency>
       <groupId>com.github.genome-nexus.genome-nexus-java-api-client</groupId>
       <artifactId>genomeNexusPublicApiClient</artifactId>
-      <version>v1.1.0-beta</version>
+      <version>5a75010393fb0a1e42275cd6ddafdf696e1ad930</version>
```
onursumer: Is this really what we want for the version number? :thinking:

leexgh: @onursumer i'm using the commit hash https://github.com/genome-nexus/genome-nexus-java-api-client/commit/5a75010393fb0a1e42275cd6ddafdf696e1ad930 .
Maybe we could create a new version?

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
leexgh: @inodb Sure will do it

inodb: we decided to wait with moving this, b/c the format of colocatedvariants has changed for vep. Better to first update that then move this logic to the server

pom.xml
inodb: @leexgh i created a new tag `v1.2.0-beta` that you can use

leexgh: @inodb thanks, already updated

leexgh:
we decide to have gnomad data from myVariantInfo

PR #88, opened 2019-03-29
Add customized columns to annotation file

leexgh:
Add two command line parameter.
1. --add-columns
Adding fields to the output file, separate fields with ","
For example:  --add-columns variantType,variantClassification

2. --prefix
Adding prefix to the header of "add-columns" fields. 
For example: --prefix gn
Example output header: gn.variantType

may need some validation for column names in the future


leexgh:
will reopen if we need this feature

PR #98, opened 2019-08-30
Functions added to order headers for MutationRecord

sheridancbio:
This combines the enhancement by @averyniceday (adding a function to give the pipelines code the same sorting / filtering functionality that was present in the "get all headers" method, plus unit tests for those function changes.

### DIFF
--- a/annotator/src/test/java/org/cbioportal/models/MutationRecordTest.java
+++ b/annotator/src/test/java/org/cbioportal/models/MutationRecordTest.java
@@ -1,0 +1,203 @@
+/*
+ * Copyright (c) 2019 Memorial Sloan-Kettering Cancer Center.
+ *
+ * This library is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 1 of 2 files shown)


annotator/src/test/java/org/cbioportal/models/MutationRecordTest.java
```diff
@@ -0,0 +1,203 @@
...
+            String valueForField = fieldName + ".value";
+            setMutationRecordField(mutationRecord, fieldName, valueForField);
+        }
+        Map<String, String> additionalPropertiesMap = new HashMap<String, String>();
+        if (additionalProperties != null) {
+            for (String additionalFieldName : additionalProperties) {
+                String valueForField = additionalFieldName + ".value";
+                additionalPropertiesMap.put(additionalFieldName, valueForField);
+            }
+
```
ao508: extra line here

PR #106, opened 2019-11-15
Add Genome Nexus version and isoform as comments at top of MAF

mandawilson:
@ao508 @sheridancbio @averyniceday is `isoformOverride` the isoform we wanted to output?

### DIFF
--- a/test/data/vcf2maf/Makefile
+++ b/test/data/vcf2maf/Makefile
@@ -21,7 +21,7 @@ test_output.uniprot.maf:
 	else \
 		AWK=awk; \
 	fi; \
-	(echo $(COMPARE_COLUMNS) | sed 's/\ /	/g'; tail -n+2 $< | $$AWK -vOFS='\t' -vFS='\t' -f ./../../scripts/select_by_column_name.awk $(COMPARE_COLUMNS)) > $@
+	(echo $(COMPARE_COLUMNS) | sed 's/\ /	/g'; grep -v "^#" $< | $$AWK -vOFS='\t' -vFS='\t' -f ./../../scripts/select_by_column_name.awk $(COMPARE_COLUMNS)) > $@
 
 all: $(ALL)
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,16 +42,16 @@ jobs:
       # run test
       - run:
           name: "check if maf file still the same when annotating with uniprot transcripts"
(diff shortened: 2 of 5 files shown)


test/data/data_mutations_extended_100.out.uniprot.txt
```diff
@@ -1,3 +1,5 @@
+#genome_nexus_version: 1.1.1-SNAPSHOT
+#isoform: uniprot
```
inodb: @mandawilson this version will prolly need to be ignored whenever one updates the version so maybe better to not have it in the test output? If you want to test inclusion of the header maybe good to create a separate test that compares the version in the header with the version in the pom instead?

test/data/vcf2maf/Makefile
```diff
@@ -21,7 +21,7 @@ test_output.uniprot.maf:
 	else \
 		AWK=awk; \
 	fi; \
-	(echo $(COMPARE_COLUMNS) | sed 's/\ /	/g'; tail -n+2 $< | $$AWK -vOFS='\t' -vFS='\t' -f ./../../scripts/select_by_column_name.awk $(COMPARE_COLUMNS)) > $@
+	(echo $(COMPARE_COLUMNS) | sed 's/\ /	/g'; grep -v "^#" $< | $$AWK -vOFS='\t' -vFS='\t' -f ./../../scripts/select_by_column_name.awk $(COMPARE_COLUMNS)) > $@
```
inodb: 🆗

test/data/data_mutations_extended_100.out.uniprot.txt
mandawilson: I see what you mean.  I just removed it but we need to then change this circle ci test: ```java -jar annotationPipeline/target/annotationPipeline-*.jar --filename test/data/data_mutations_extended_100.in.txt  --output-filename test/data/data_mutations_extended_100.out.uniprot.txt --isoform-override uniprot && git diff --exit-code test/data/data_mutations_extended_100.out.uniprot.txt || (echo -e "MAF uniprot output changed test/data/data_mutations_extended_100.out.uniprot.txt $(cat test/data/data_mutations_extended_100.out.uniprot.txt | curl -F c=@- https://ptpb.pw | grep url) (include this file in your PR to fix this test)" && exit 1)``` We should probably write to a different file, and then do an ordinary diff that cuts out any comments.  What do you think @inodb?

mandawilson: @inodb Actually I think I can change the git diff, and then it should work.

PR #108, opened 2019-11-19
Use POST requests for GN annotation pipeline

ao508:
Use POST requests for Genome Nexus annotation pipeline

Added option to run the annotation pipeline using POSTs instead of GETs when calling Genome Nexus web service.
- To enable POST requests: Add `-p <number>` to the command line arguments where `<number>` is the chunk/POST request interval size to send to GN.

- Added `AnnotationSummaryStatistics` to keep track of failed annotations, error messages, and saves error report to file if desired.

- General refactoring and cleanup of `MutationRecordReader` and `GenomeNexusImpl`.
- Removed `verbose` option from command line.


annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -73,12 +75,7 @@
     private String enrichmentFields;
 
     private AnnotationControllerApi apiClient;
-
-    private MutationRecord mRecord;
-    private VariantAnnotation gnResponse;
```
mandawilson: `mRecord`, `gnResponse`, and `canonicalTranscript` are no longer class variables, instead passed around as parameters to the methods.

annotationPipeline/src/main/java/org/cbioportal/annotation/pipeline/MutationRecordReader.java
```diff
@@ -157,43 +199,25 @@ public void handleLine(String line) {
...
-                else {
-                    annotationErrorMessage = "Failed to annotate variant";                    
-                }
-                failedAnnotations++;
-            }
-            if (!annotationErrorMessage.isEmpty()) {
-                if (verbose) LOG.info(annotationErrorMessage + ": " + variantDetails);
-                if (errorReportLocation != null) updateErrorMessages(record, annotatedRecord.getVARIANT_CLASSIFICATION(), annotator.getUrlForRecord(record, isoformOverride), annotationErrorMessage);                
-            }
+            // dont need to do anything with output, just need to call method
```
mandawilson: Great.

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -532,4 +534,35 @@ private TranscriptConsequenceSummary getCanonicalTranscript(VariantAnnotation gn
         hgvspNullClassifications.add("RNA");
         return hgvspNullClassifications;
     }
+
+    @Override
+    public List<AnnotatedRecord> getAnnotatedRecordsUsingPOST(AnnotationSummaryStatistics summaryStatistics, List<MutationRecord> mutationRecords, String isoformOverridesSource, Boolean replace) throws Exception {
+        // construct list of genomic location objects to pass to api client
+        List<GenomicLocation> genomicLocations = new ArrayList<>();
```
mandawilson: Since we are creating a new list of `GenomicLocations` to pass to the annotator, really this method could just take the original (unpartitioned) list of `MutationRecords` and a start and end index as parameters.  Your way is probably clearer, but is creating a lot of unnecessary lists.  Probably clarity of code is more important than any tiny gain in memory/efficiency.  Yeah I like your way best.

annotationPipeline/src/main/java/org/cbioportal/annotation/pipeline/MutationRecordReader.java
```diff
@@ -92,11 +82,30 @@
...
         String genomeNexusVersion = annotator.getVersion();
 
         processComments(ec, genomeNexusVersion);
+        List<MutationRecord> mutationRecords = loadMutationRecordsFromMaf();
+        if (postIntervalSize > 0) {
+            try {
+                this.allAnnotatedRecords = annotateRecordsWithPOST(mutationRecords);
+            } catch (Exception ex) {
+                LOG.error("ERROR ANNOTATING WITH POST REQUESTS");
+                throw new RuntimeException(ex);
```
mandawilson: I guess here we throw an exception because we are losing lots of records at once.  In GET, below, we don't throw an exception but rather log the error.  Are we sure we want to handle them differently?  I think probably yes.

ao508:
@inodb 

Here are the runtimes when annotating the full MSKIMPACT MAF with GETs vs. POSTs.

- GETs: ~3.5 hours
- POSTs: ~15 minutes

The MAF contains about 350k records and the annotator was using our pipelines VEP server on AWS.

n1zea144:
This is great -  our curators will really like this improvement!

inodb:
@ao508 Great speed-up 🎉 !

Some questions:

- are the resulting MAFs identical?
- I assume this is when everything is cached. Maybe good to check without caching as well?

For the tests:

- I think we could e.g. duplicate these lines: https://github.com/genome-nexus/genome-nexus-annotation-pipeline/blob/master/.circleci/config.yml#L43-L58 but then run with the POST parameter. The results should be identical

ao508:
@inodb 

> Some questions:
> are the resulting MAFs identical?

There are some discrepancies that I found in the `Variant_Classification` column. The annotated MAF has a mixture of non-normalized such as "frameshift_insertion" instead of its normalized value of "Frame_Shift_Ins".

> I assume this is when everything is cached. Maybe good to check without caching as well?

Sure, I can try this out as well while I dig into the above issue.

> For the tests:
>
> I think we could e.g. duplicate these lines: https://github.com/genome-nexus/genome-nexus-annotation-pipeline/blob/master/.circleci/config.yml#L43-L58 but then run with the POST parameter. The results should be identical

I'll dig into this as well while I figure out that issue above.

ao508:
@inodb  the discrepancies I found in the `Variant_Classification` column were resolved by a minor update to the `VariantClassificationResolver` in the server code.

ao508:
@inodb @mandawilson @sheridancbio Updated this PR with unit tests.

inodb:
@ao508 Nice work 🚀 !

For the server update to fix the `VariantClassificationResolver` - is there a PR for that to genome-nexus?

Not really necessary for this PR, but maybe nice to try against grch37.rest.ensembl.org as well without cache? That'll give an idea of whether the EC2 VEP performs similar to the live ensembl web service

inodb:
Minor thing re tests: I don't see these lines copied and using POST instead https://github.com/genome-nexus/genome-nexus-annotation-pipeline/blob/master/.circleci/config.yml#L43-L58? Can we add those? I think it's useful b/c it covers all vcf2maf test cases as well

ao508:
@inodb do you know what this error is from circleci ? I tried just refreshing the PR to get the build going again but got the same error:

```
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0curl: (7) Failed to connect to ptpb.pw port 443: Connection refused

```

inodb:
@ao508 ah that ptpb.pw service is down. I used to use that one to upload the maf file in case of failed tests. That part can be removed. Alternatively we could use CircleCI Artifacts and store the file there

PR #110, opened 2019-12-02
Adjust integration test results

sheridancbio:
- our production genome nexus server appears to have drifted from the original expected annotations and is now giving slightly modified annotations in some instances. We are not sure why this change has happened, but we don't see any obvious mistakes in the new annotations, so we are updating the expected output for the unit tests.


sheridancbio:
@inodb : as you suggested, we ran the annotation script on the current master branch and the same differences in annotation are observed. So we have have adjusted the expected output for now. In general, this is a drawback to testing against a live production system rather than a mock system. I think moving towards a more isolated mock system would be a better approach.

PR #116, opened 2020-01-03
Catch GN annotation failure exceptions

ao508:


ao508:
@averyniceday I dont see what you're referring to

PR #123, opened 2020-04-02
Increase read timeout for ApiClient

ao508:
Resolves `SocketTimeout Exception`'s observed when reading responses from VEP/ENSEMBL REST service. Note that this exception is also observed even when running VEP locally (through VEP wrapper). 

This PR replaces the pull request created in the [genome-nexus-java-api-client](https://github.com/genome-nexus/genome-nexus-java-api-client/pull/5) repo.

### DIFF
--- a/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
+++ b/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
@@ -83,6 +83,7 @@ public class GenomeNexusImpl implements Annotator {
 
     private final Pattern PROTEIN_POSITTION_REGEX = Pattern.compile("p.[A-Za-z]([0-9]*).*$");
     private static List<String> hgvspNullClassifications = initNullClassifications();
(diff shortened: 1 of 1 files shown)


ao508:
@onursumer I closed the PR in the `genome-nexus-java-api-client` repo and replaced it with this one. Thanks again for your help!

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -202,6 +203,7 @@ private AnnotationControllerApi initApiClient()
         if (this.genomeNexusBaseUrl != null && this.genomeNexusBaseUrl.length() > 0)
         {
             ApiClient client = new ApiClient();
+            client.setReadTimeout(300);
```
onursumer, line 206: aren't we supposed to use `READ_TIMEOUT_OVERRIDE ` here? Also should it be `300` or `300000`?

ao508: @onursumer lol thanks, updating now

PR #125, opened 2020-04-13
Resolve DB SNP RS id value

ao508:


annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -703,6 +703,15 @@ private TranscriptConsequenceSummary getCanonicalTranscript(VariantAnnotation gn
         }
     }
 
+    private String resolveDbSnpRs(VariantAnnotation gnResponse, MutationRecord mRecord) {
+        String dbSnpRs = null;
+        if (gnResponse.getColocatedVariants() != null && !gnResponse.getColocatedVariants().isEmpty()) {
+            dbSnpRs = gnResponse.getColocatedVariants().get(0).getDbSnpId();
```
sheridancbio, line 709: I guess the colocated variants are a collection/list? If the list contains rs0000001 and rs0000002, is it correct to return the first one? I would guess that these would be distinct variations which occur at the some location. So maybe rs0000001 is an A->C substitution and rs0000002 is an A->G substitution at the same genomic position. Would we need to look up the nucleotide variation involved in these two entries and only return the record that matches the query? (perhaps neither would match the query ... if the query was for an A->T substitution ... so then perhaps we should return nothing?)
I don't know the answer ... just raising the question. Is there any related information included in the output regarding colocated variants? There is a code called univar_id which would tell us the base substitution to see if we matched.

ao508: @sheridancbio we assume the first one

ao508:
@sheridancbio I vaguely recall seeing a conversation with you and @inodb in the slack regarding the results in the "co-located" variants field? I'm having trouble finding it now but if either of you would mind sharing some details of your conversation here that would be appreciated. I seem to remember that there can be more than just db SNP IDs in the response?

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -703,6 +705,20 @@ private TranscriptConsequenceSummary getCanonicalTranscript(VariantAnnotation gn
         }
     }
 
+    private String resolveDbSnpRs(VariantAnnotation gnResponse, MutationRecord mRecord) {
+        String dbSnpRs = null;
+        if (gnResponse.getColocatedVariants() != null && !gnResponse.getColocatedVariants().isEmpty()) {
+            for (ColocatedVariant cv : gnResponse.getColocatedVariants()) {
+                Matcher dbSnpRsIdMatcher = DBSNP_RSID_REGIX.matcher(cv.getDbSnpId());
+                if (dbSnpRsIdMatcher.find()) {
+                    dbSnpRs = dbSnpRsIdMatcher.group(1);
```
averyniceday, line 714: Could we use the whole match? Or is there a reason we look for specifically the first group?

ao508: @averyniceday updated to whole match

ao508:
@inodb have you encountered this curl timeout issue in the circleci run before? I've been trying to trace it but I cant recreate the error locally. [see here](https://circleci.com/gh/genome-nexus/genome-nexus-annotation-pipeline/230?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

PR #126, opened 2020-04-17
Reintroduce Re-annotation Flag

averyniceday:


annotationPipeline/src/main/java/org/cbioportal/annotation/pipeline/MutationRecordReader.java
```diff
@@ -85,9 +85,9 @@ public void open(ExecutionContext ec) throws ItemStreamException {
         processComments(ec, genomeNexusVersion);
         List<MutationRecord> mutationRecords = loadMutationRecordsFromMaf();
         if (postIntervalSize > 0) {
-            this.allAnnotatedRecords = annotator.getAnnotatedRecordsUsingPOST(summaryStatistics, mutationRecords, isoformOverride, replace, postIntervalSize);
+            this.allAnnotatedRecords = annotator.getAnnotatedRecordsUsingPOST(summaryStatistics, mutationRecords, isoformOverride, replace, postIntervalSize, true);
```
averyniceday, line 88: This is preserving the ORIGINAL logic which was accidentally lost. We always wanted to re-annotate when we run genome-nexus-annotation-pipeline jar from the command line.
Genome nexus annotation pipeline does not have a command line flag for forcing re-annotation. Instead it has the replace flag for replacing HUGO and Entrez gene Ids.

Logic here:
https://github.com/genome-nexus/genome-nexus-annotation-pipeline/pull/108/files#diff-3d8ddf8f301d85f3990b5327a09a79f7R188

annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -716,19 +716,21 @@ private TranscriptConsequenceSummary getCanonicalTranscript(VariantAnnotation gn
...
     @Override
-    public List<AnnotatedRecord> getAnnotatedRecordsUsingPOST(AnnotationSummaryStatistics summaryStatistics, List<MutationRecord> mutationRecords, String isoformOverridesSource, Boolean replace, Integer postIntervalSize) {
+    public List<AnnotatedRecord> getAnnotatedRecordsUsingPOST(AnnotationSummaryStatistics summaryStatistics, List<MutationRecord> mutationRecords, String isoformOverridesSource, Boolean replace, Integer postIntervalSize, boolean reannotate) {
         // construct list of genomic location objects to pass to api client
         // TODO use SortedSet (or at least Set) instead?  Do we anticipate a lot of redundancy? Probably quite a bit.
         // Maybe test which is faster with a large study
         List<GenomicLocation> genomicLocations = new ArrayList<>();
         for (MutationRecord record : mutationRecords) {
-            genomicLocations.add(extractGenomicLocation(record));
+            if(reannotate || annotationNeeded(record)) {
```
sheridancbio, line 731: Normally I would recommend a space after the keyword 'if', but this is a minor point and we wish to deploy this in short order.

PR #129, opened 2020-04-29
Prevent annotation when no records extracted from input file.

ao508:
If there aren't any records loaded from the input MAF then the annotation pipeline does not attempt to submit requests to Genome Nexus. 

The count of records loaded/annotated is added to the execution context so that the `MutationRecordWriter` can use the value to determine if the output file needs to be generated or not. 


averyniceday:
Seems like this needs to be rebased onto master

PR #131, opened 2020-05-04
Instructions for Updating GN Annotation Pipeline

averyniceday:
Update documentation with instructions for updating models.
Specific instructions also added in conjunction to Genome Nexus Java api client codebase. 

GN Java API Client PR here: https://github.com/genome-nexus/genome-nexus-java-api-client/pull/6

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,13 @@
 These tools allow for annotation of genomic variants from a MAF for import into
 the cBioPortal using [Genome Nexus](http://genomenexus.org)
 
+## Updating the Genome Nexus Annotation Pipeline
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,6 +2,13 @@
 These tools allow for annotation of genomic variants from a MAF for import into
 the cBioPortal using [Genome Nexus](http://genomenexus.org)
 
+## Updating the Genome Nexus Annotation Pipeline
```
sheridancbio, line 5: I think this section should move down to the bottom of this document. The other sections tell "What tools" are referred to by the opening section above, so I would not split that.
To emphasize that there is guidance on how to make updates, a table of contents could be added up top with links to the various sections.

PR #132, opened 2020-05-26
Add 'Annotation_Status' field

ao508:
Added a new field called "Annotation_Status" that will store "SUCCESS" or "FAILED" to indicate if an annotation attempt to Genome Nexus succeeded. Even if the variant falls in a non-coding region, if the annotation attempt was still technically successful then this field will contain "SUCCESS". "FAILED" will only be used when an error occurs on the server side. 


annotator/src/main/java/org/cbioportal/models/AnnotatedRecord.java
```diff
@@ -396,6 +397,14 @@ public void setGNOMAD_SAS_AF(String gnomadAlleleFrequencySAS) {
         this.gnomadAlleleFrequencySAS = gnomadAlleleFrequencySAS;
     }
 
+    public String getANNOTATION_STATUS() {
+        return this.annotationStatus != null ? this.annotationStatus : "FAILED";
```
sheridancbio, line 401: wouldn't it be cleaner to always set the annotationStatus when we attempt annotation? (see my comment above for the HTTP GET) ... if this.annotationStatus is null, that could be interpreted as meaning "we did not attempt to annotated this variant" rather than assuming it means the annotation failed. By interpreting null as meaning FAILED, we lose that distinction.

PR #148, opened 2020-09-22
Support POLYPHEN, SIFT annotations in output annotated MAF

ao508:
TO-DO:

-  update pom commit hash and group id when changes to the java api
client are merged


sheridancbio (approved):
👍

PR #156, opened 2020-12-07
PCM1 DNP annotation issue

inodb:
I've added several test cases in this PR. The main issue that this addresses is that Tumor_Seq_Allele1 wouldn't be updated when the reference allele was changed, even though in the input the mutations were identical.

Basically this:

```
8	17796381	17796383	AAC	AAC	AGT
```
became this after annotating
```
8	17796382	17796383	AC	AAC	GT
```
Now it's fixed to output:
```
8	17796382	17796383	AC	AC	GT
```

I do have my doubts about changing/normalizing the mutations in general. It would be better to treat chrom,start,end,ref,alt as immutable imo. We've had multiple issues related to this, so maybe we should turn this off completely? If it's essential for particular studies, maybe we can opt-in with some command line flag (EDIT: filed an issue https://github.com/genome-nexus/genome-nexus-annotation-pipeline/issues/157)


annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java
```diff
@@ -213,6 +213,24 @@ public AnnotatedRecord convertResponseToAnnotatedRecord(VariantAnnotation gnResp
...
+        if (mRecord.getTUMOR_SEQ_ALLELE1().equals(mRecord.getREFERENCE_ALLELE())) {
+            resolvedTumorSeqAllele1 = resolvedReferenceAllele;
+        } else if (mRecord.getTUMOR_SEQ_ALLELE1().equals(mRecord.getTUMOR_SEQ_ALLELE2())) {
+            resolvedTumorSeqAllele1 = resolvedTumorSeqAllele2;
+        } else {
+            // TODO: it's also possible that the position has changed after
+            // resolving ref+alt. Tumor seq allele1 would have to be updated
+            // then as well. Kind of a corner case, but we might want to handle
+            // that in some way. Discussion point: should we even allow the core
+            // variant attributes (pos,ref,alt1,alt2) to be mutable?
```
ao508, line 231: This is the question of the last few years. It's the same behavior that vcf2maf/maf2maf supported right? If we don't alter the position(s) then the annotation for the variant would fail based on VEP's reference genome sequence for the input variant's coordinates not matching the reference allele that we are submitting for annotation. 

So do we continue our current approach and allow the core variant attributes to be mutable or do we fail those annotations and require the user to fix and/or adjust them themselves?

ao508:
@sheridancbio @mandawilson @averyniceday Jenkins isn't able to run this workflow properly. I think I've narrowed down the issue to the script that checks for whether the owner of the PR is a trusted user or not. 

I ran the script manually as the jenkins user and this is the response I got 

```
{u'documentation_url': u'https://docs.github.com/articles/creating-a-personal-access-token-for-the-command-line', u'message': u"Bad credentials. The API can't be accessed using username/password authentication. Please create a personal access token to access this endpoint: http://github.com/settings/tokens"}
```


I see a token had already been generated for the jenkins user but I think we need to update this `check-pr-from-trusted-users.sh` script to use the token instead of its git credentials file. 

Would one of you be able to assist with this?

ao508 (commented):
I think the code looks good, I would like to see the jenkins tests succeed though first before merging. I've asked rob, manda, and avery for their assistance with this task.

sheridancbio:
@ao508 @inodb we are still working on the block for the jenkins tests, but we bypassed the check temporarily and the rest of the jenkins job ran to completion (successfully).

PR #158, opened 2020-12-15
Add gnomad annotation test

inodb:
This was a case from Tom Yo that didn't annotate with gnomad previously


inodb:
just gonna merge this b/c I believe the jenkins job is failing for a diff reason

PR #161, opened 2021-02-04
Use core module only

inodb:
Using the core module instead of the entire portal package somehow
solves the problem. That did lack some deps such as log4j and mysql. A
quick hack was to just include the log4j dep from cbioportal. Instead of
the mysql package, I think we can use vanilla sql according to this
stackoverflow answer: https://stackoverflow.com/a/27582891/1894184.

WARNING: we tried to include only the core module before and that caused some issue for a pipeline package downstream (see https://github.com/genome-nexus/genome-nexus-annotation-pipeline/pull/107). That might still be an issue. Maybe we can figure out a way to change these downstream deps to not depend on the entire portal package CC @averyniceday ?

### DIFF
--- a/databaseAnnotator/src/main/java/org/cbioportal/database/annotator/AnnotateRecordsWriter.java
+++ b/databaseAnnotator/src/main/java/org/cbioportal/database/annotator/AnnotateRecordsWriter.java
@@ -32,7 +32,6 @@
 
 package org.cbioportal.database.annotator;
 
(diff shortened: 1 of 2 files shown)


ao508:
@inodb I confirmed that the CVR fetcher works with your GNAP changes here. I may want to run my own testing with the importers and standalone annotation pipeline first before moving forward with this PR. I will keep you posted.

ao508 (approved):

PR #163, opened 2021-02-06
Handle changes to deps in cbioportal v3.6.4

inodb:
Somehow the old logging configuration won't work anymore (getting some
logconfiguration error). Made some changes to the logging framework.
This does quite drastically change output since everything now goes to
stdout for some reason. Maybe someone with more java logging experience
can figure out what's really going on, but this could work as a quick
fix.


ao508:
Ah, this is why we were excluding the log4j stuff from the cbioportal package

inodb:
@ao508 i guess so yeah - i made some changes here: https://github.com/genome-nexus/genome-nexus-annotation-pipeline/pull/165. Which seems less invasive to the logging at least

PR #165, opened 2021-02-12
Handle changes to deps in cbioportal v3.6.4

inodb:
Somehow the old logging configuration won't work anymore (getting some
logconfiguration error). This fixes that by reverting some changes to
loggin framework exclusion. In addition the the spring boot version was
upgraded. Peculiarly some in memory database had to be included. Not
entirely sure how the dependency tree works, but it is probably due to
changes in the databaseAnnotator configuration in the new spring boot
version.

### DIFF
--- a/annotationPipeline/pom.xml
+++ b/annotationPipeline/pom.xml
@@ -58,4 +58,14 @@
     </dependency>
   </dependencies>
 
+  <build>
+    <plugins>
+      <plugin>
+        <!-- required to build an executable jar -->
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-maven-plugin</artifactId>
+      </plugin>
+    </plugins>
+  </build>
+
 </project>
--- a/annotationPipeline/src/main/java/org/cbioportal/annotation/AnnotationPipeline.java
+++ b/annotationPipeline/src/main/java/org/cbioportal/annotation/AnnotationPipeline.java
@@ -35,8 +35,8 @@ package org.cbioportal.annotation;
 import org.cbioportal.annotation.pipeline.BatchConfiguration;
 
 import org.apache.commons.cli.*;
-import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.builder.SpringApplicationBuilder;
 import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.batch.core.*;
 import org.springframework.batch.core.launch.JobLauncher;
@@ -72,8 +72,7 @@ public class AnnotationPipeline
     private static void launchJob(String[] args, String filename, String outputFilename, String isoformOverride,
             String errorReportLocation, boolean replace, Integer postIntervalSize) throws Exception
     {
-        SpringApplication app = new SpringApplication(AnnotationPipeline.class);
-        ConfigurableApplicationContext ctx = app.run(args);
+        ConfigurableApplicationContext ctx = new SpringApplicationBuilder(AnnotationPipeline.class).web(false).run(args);
         JobLauncher jobLauncher = ctx.getBean(JobLauncher.class);
 
         Job annotationJob = ctx.getBean(BatchConfiguration.ANNOTATION_JOB, Job.class);
--- a/annotator/pom.xml
+++ b/annotator/pom.xml
@@ -33,6 +33,29 @@
       <artifactId>commons-cli</artifactId>
       <version>1.3</version>
     </dependency>
+    <dependency>
+        <groupId>javax.xml.bind</groupId>
+         <artifactId>jaxb-api</artifactId>
+        <version>2.3.0</version>
+     </dependency>
+    <dependency>
+        <groupId>com.h2database</groupId>
+        <artifactId>h2</artifactId>
+        <version>1.3.156</version>
+    </dependency>
   </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <!-- required to build an executable jar -->
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-maven-plugin</artifactId>
+        <configuration>
+            <skip>true</skip>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
 
 </project>
--- a/databaseAnnotator/pom.xml
+++ b/databaseAnnotator/pom.xml
@@ -118,4 +118,14 @@
     </repository>
   </repositories>
 
+  <build>
+    <plugins>
+      <plugin>
+        <!-- required to build an executable jar -->
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-maven-plugin</artifactId>
+      </plugin>
+    </plugins>
+  </build>
+
 </project>
--- a/databaseAnnotator/src/main/java/org/cbioportal/database/DatabaseAnnotator.java
+++ b/databaseAnnotator/src/main/java/org/cbioportal/database/DatabaseAnnotator.java
@@ -35,8 +35,8 @@ package org.cbioportal.database;
 import org.cbioportal.database.annotator.BatchConfiguration;
 import org.apache.commons.cli.*;
 import org.apache.commons.cli.Options;
-import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.builder.SpringApplicationBuilder;
 import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.batch.core.*;
 import org.springframework.batch.core.launch.JobLauncher;
@@ -50,9 +50,9 @@ import org.apache.log4j.Logger;
 @SpringBootApplication
 public class DatabaseAnnotator {
 
-    private static Logger log = Logger.getLogger(DatabaseAnnotator.class);
+    private static Logger LOG = Logger.getLogger(DatabaseAnnotator.class);
 
-    private static Options getOptions(String[] args){
+    private static Options getOptions(String[] args) {
         Options gnuOptions = new Options();
         gnuOptions.addOption("h", "help", false, "shows this help document and quits.");
         gnuOptions.addOption("i", "isoform", true, "Isoform desired for annotation. mskcc or uniprot");
@@ -60,28 +60,25 @@ public class DatabaseAnnotator {
         return gnuOptions;
     }
 
-    private static void help(Options gnuOptions, int exitStatus){
+    private static void help(Options gnuOptions, int exitStatus) {
         HelpFormatter helpFormatter = new HelpFormatter();
         helpFormatter.printHelp("Darwin Pipeline", gnuOptions);
         System.exit(exitStatus);
     }
 
+    private static void launchJob(String[] args, String isoform, String studies) throws Exception {
+        ConfigurableApplicationContext ctx = new SpringApplicationBuilder(DatabaseAnnotator.class).web(false).run(args);
 
-    private static void launchJob(String[] args, String isoform, String studies) throws Exception{
-        SpringApplication app = new SpringApplication(DatabaseAnnotator.class);
-
-        ConfigurableApplicationContext ctx = app.run(args);
         JobLauncher jobLauncher = ctx.getBean(JobLauncher.class);
-        JobParameters jobParameters = new JobParametersBuilder()
-                .addString("isoform", isoform)
-                .addString("studies", studies)
-                .toJobParameters();
+        JobParameters jobParameters = new JobParametersBuilder().addString("isoform", isoform)
+                .addString("studies", studies).toJobParameters();
+
         Job databaseAnnotatorJob = ctx.getBean(BatchConfiguration.DATABASE_ANNOTATOR_JOB, Job.class);
         JobExecution jobExecution = jobLauncher.run(databaseAnnotatorJob, jobParameters);
         ctx.close();
     }
 
-    public static void main(String[] args) throws Exception{
+    public static void main(String[] args) throws Exception {
         Options gnuOptions = DatabaseAnnotator.getOptions(args);
         CommandLineParser parser = new GnuParser();
         CommandLine commandLine = parser.parse(gnuOptions, args);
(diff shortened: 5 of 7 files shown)


annotationPipeline/src/main/java/org/cbioportal/annotation/AnnotationPipeline.java
```diff
@@ -72,8 +72,7 @@ private static void help(Options gnuOptions, int exitStatus)
     private static void launchJob(String[] args, String filename, String outputFilename, String isoformOverride,
             String errorReportLocation, boolean replace, Integer postIntervalSize) throws Exception
     {
-        SpringApplication app = new SpringApplication(AnnotationPipeline.class);
-        ConfigurableApplicationContext ctx = app.run(args);
+        ConfigurableApplicationContext ctx = new SpringApplicationBuilder(AnnotationPipeline.class).web(false).run(args);
```
ao508, line 75: There’s a spring property that you can add to disable the default web startup. I’m on mobile right now otherwise I’d share but if you don’t mind waiting for Monday I can share it then

inodb: sure - thanks! This can def wait: I do wonder if we ever want to start up web? For the command line client it might make sense to disable completely?

pom.xml
```diff
@@ -69,7 +69,13 @@
...
       <groupId>com.github.cbioportal.cbioportal</groupId>
 	  <artifactId>core</artifactId>
-      <version>v3.6.2</version>
+      <version>24bc8b56d73cacb33a1601e3a60160ceff2fa13d</version>
+      <exclusions>
+          <exclusion>
+              <groupId>org.apache.logging.log4j</groupId>
+              <artifactId>log4j-core</artifactId>
+          </exclusion>
+      </exclusions>
```
inodb, line 78: this was excluded before, but i included it when we made this a core only dep - excluding it again

pom.xml
```diff
@@ -19,7 +19,7 @@
   <parent>
     <groupId>org.springframework.boot</groupId>
     <artifactId>spring-boot-starter-parent</artifactId>
-    <version>1.2.7.RELEASE</version>
+    <version>1.4.7.RELEASE</version>
```
inodb, line 22: figured i might use this opportunity to also upgrade spring a bit (higher versions give some problems with the mocking class atm)

databaseAnnotator/src/main/java/org/cbioportal/database/annotator/DataSourceConfiguration.java
```diff
@@ -61,7 +60,7 @@
     @Value("${databaseannotator.connection_string}")
     private String connection_string;
 
-    @Bean
+    @Bean(destroyMethod = "")
```
inodb, line 63: this was giving some jmx warning without using destroyMethod

databaseAnnotator/src/main/java/org/cbioportal/database/DatabaseAnnotator.java
```diff
@@ -50,38 +50,35 @@
...
-        SpringApplication app = new SpringApplication(DatabaseAnnotator.class);
-
-        ConfigurableApplicationContext ctx = app.run(args);
         JobLauncher jobLauncher = ctx.getBean(JobLauncher.class);
-        JobParameters jobParameters = new JobParametersBuilder()
-                .addString("isoform", isoform)
-                .addString("studies", studies)
-                .toJobParameters();
+        JobParameters jobParameters = new JobParametersBuilder().addString("isoform", isoform)
+                .addString("studies", studies).toJobParameters();
```
inodb, line 74: IDE reformatting 🤷‍♂️

databaseAnnotator/src/main/java/org/cbioportal/database/DatabaseAnnotator.java
```diff
@@ -50,38 +50,35 @@
...
 
-    private static void help(Options gnuOptions, int exitStatus){
+    private static void help(Options gnuOptions, int exitStatus) {
         HelpFormatter helpFormatter = new HelpFormatter();
         helpFormatter.printHelp("Darwin Pipeline", gnuOptions);
         System.exit(exitStatus);
     }
 
+    private static void launchJob(String[] args, String isoform, String studies) throws Exception {
+        ConfigurableApplicationContext ctx = new SpringApplicationBuilder(DatabaseAnnotator.class).web(false).run(args);
```
inodb, line 70: this is new, disabling web also for the database annotator (not sure if this even still works, the databaseannotator i mean)

inodb:
Also created two PRs to see if this would break downstream pipeline inclusions:

- https://github.com/knowledgesystems/cmo-pipelines/pull/828
- https://github.com/knowledgesystems/pipelines/pull/413

ao508:
looking at this now

annotationPipeline/src/main/java/org/cbioportal/annotation/AnnotationPipeline.java
ao508: can you add this to the `application.properties.EXAMPLE` and your app properties? 

```
# required to prevent embedded spring boot web server from starting up
spring.main.web-application-type=none
```

and remove the change introduced here?

annotator/pom.xml
```diff
@@ -33,6 +33,29 @@
...
 
+  <build>
+    <plugins>
+      <plugin>
+        <!-- required to build an executable jar -->
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-maven-plugin</artifactId>
+        <configuration>
+            <skip>true</skip>
+        </configuration>
```
inodb, line 56: this was giving some error with the spring boot upgrade

annotator/pom.xml
```diff
@@ -33,6 +33,29 @@
...
   </dependencies>
 
+  <build>
+    <plugins>
+      <plugin>
+        <!-- required to build an executable jar -->
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-maven-plugin</artifactId>
+        <configuration>
+            <skip>true</skip>
```
ao508, line 55: 👍

annotator/pom.xml
```diff
@@ -33,6 +33,29 @@
...
+    <dependency>
+        <groupId>javax.xml.bind</groupId>
+         <artifactId>jaxb-api</artifactId>
+        <version>2.3.0</version>
+     </dependency>
+    <dependency>
+        <groupId>com.h2database</groupId>
+        <artifactId>h2</artifactId>
+        <version>1.3.156</version>
+    </dependency>
```
inodb, line 45: needed some database, otherwise there'd be some errors about db configuration. I haven't really tested the database annotator. Not sure if we are still using that?

annotator/pom.xml
```diff
@@ -33,6 +33,29 @@
       <artifactId>commons-cli</artifactId>
       <version>1.3</version>
     </dependency>
+    <dependency>
+        <groupId>javax.xml.bind</groupId>
+         <artifactId>jaxb-api</artifactId>
+        <version>2.3.0</version>
```
inodb, line 39: i think because further down i disabled the configuration of this module using spring boot I needed to add this dependency

pom.xml
ao508: I am a little curious about why we need to explicitly disable the start up web context suddenly if it wasn't necessary for this project before. 

I suspect it's related to the updated spring boot version?

The `cmo-pipelines` is using `1.2.7` still so if the version updates here then I suspect there'll be some conflicts to resolve in `cmo-pipelines` as well. Is this upgrade required here or can it wiat?

databaseAnnotator/src/main/java/org/cbioportal/database/DatabaseAnnotator.java
```diff
@@ -50,38 +50,35 @@
 @SpringBootApplication
 public class DatabaseAnnotator {
 
-    private static Logger log = Logger.getLogger(DatabaseAnnotator.class);
+    private static Logger LOG = Logger.getLogger(DatabaseAnnotator.class);
```
inodb, line 53: In some places we're using `LOG`, in others `log`. It seems like we're using `LOG` more often

annotationPipeline/pom.xml
```diff
@@ -58,4 +58,14 @@
...
 
+  <build>
+    <plugins>
+      <plugin>
+        <!-- required to build an executable jar -->
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-maven-plugin</artifactId>
+      </plugin>
+    </plugins>
+  </build>
```
inodb, line 69: Building the jar might work without adding the plugin in this module (it's already in the parent). It doesn't hurt to add it though afaik, just makes it easier to configure at the module level in the future if we want. It was only really necessary for the `annotator` module b/c that one would no longer work with the spring boot upgrade

databaseAnnotator/pom.xml
```diff
@@ -118,4 +118,14 @@
...
 
+  <build>
+    <plugins>
+      <plugin>
+        <!-- required to build an executable jar -->
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-maven-plugin</artifactId>
+      </plugin>
+    </plugins>
+  </build>
```
inodb, line 129: Building the jar might work without adding the plugin in this module (it's already in the parent). It doesn't hurt to add it though afaik, just makes it easier to configure at the module level in the future if we want. It was only really necessary for the `annotator` module b/c that one would no longer work with the spring boot upgrade

ao508 (commented):
made a couple comments

annotationPipeline/src/main/java/org/cbioportal/annotation/AnnotationPipeline.java
inodb: @ao508 i think in this case it might make more sense to hardcode it. E.g. if someone used a previous `properties.EXAMPLE` file to set up their annotator and they'd try to upgrade they would get a very hard to understand error. They would then have to figure out they have to make changes to their properties file. In addition it seems to me like we always want to avoid the embedded spring boot web server, so why would we make it configurable?

inodb: If we had some `defaults.properties` that gets overridden by `application.properties` then we could add that configuration there. That might be cleaner, but not sure if there's an easy way to set that up? I don't think `.EXAMPLE` is currently configured to work that way?

pom.xml
inodb: Ah I see - good point. I'll see if i can split out the spring upgrade

inodb: @ao508 so yeah looks like that doesn't work b/c somehow the log configuration then breaks:

`Caused by: java.lang.NoClassDefFoundError: org/springframework/util/Log4jConfigurer`

I guess by upgrading spring boot it is able to use log4j from that one 🤔

inodb: see e.g. the dep tree:
```
[INFO] +- org.springframework.boot:spring-boot-starter-batch:jar:1.4.7.RELEASE:compile
[INFO] |  +- org.springframework.boot:spring-boot-starter:jar:1.4.7.RELEASE:compile
[INFO] |  |  +- org.springframework.boot:spring-boot:jar:1.4.7.RELEASE:compile
[INFO] |  |  +- org.springframework.boot:spring-boot-autoconfigure:jar:1.4.7.RELEASE:compile
[INFO] |  |  +- org.springframework.boot:spring-boot-starter-logging:jar:1.4.7.RELEASE:compile
[INFO] |  |  |  +- org.slf4j:jcl-over-slf4j:jar:1.7.30:compile
[INFO] |  |  |  \- org.slf4j:jul-to-slf4j:jar:1.7.30:compile
```

annotationPipeline/src/main/java/org/cbioportal/annotation/AnnotationPipeline.java
ao508: no it isn't, it's just meant to be there as an example properties file with default values. I guess we could set the other default props to point to the public genome nexus and if `application.properties` doesn't exist then the mvn build can copy the example props file similar to how it's done in the `cbioportal` project.

pom.xml
ao508: in `cmo-pipelines` we are also excluding  some logging dependencies from the spring boot starter batch dep


```
    <dependency>
      <groupId>org.springframework.boot</groupId>
      <artifactId>spring-boot-starter-batch</artifactId>
      <exclusions>
        <exclusion>
          <groupId>org.slf4j</groupId>
          <artifactId>log4j-over-slf4j</artifactId>
        </exclusion>
        <exclusion>
          <groupId>ch.qos.logback</groupId>
          <artifactId>logback-classic</artifactId>
        </exclusion>
      </exclusions>
    </dependency>
```

does it end up working if we do the same for this project while keeping the version 1.2.7?

databaseAnnotator/pom.xml
ao508: if we're adding the plugin to all modules then maybe we should remove it from the parent?

ao508: If we remove it from the parent then I guess we wouldnt need this to be skipped then in the `annotator` module, right?

pom.xml
inodb: @ao508 that's actually already excluded here: 

https://github.com/genome-nexus/genome-nexus-annotation-pipeline/blob/09b0a0522d04b7b2bffa94c3b8c7df7f67da4a5f/pom.xml#L36-L46

> I suspect it's related to the updated spring boot version?

It's actually not related to spring boot version update in this PR turns out but rather to the dependency management changes in this PR: https://github.com/cBioPortal/cbioportal/pull/7984

inodb (commented):
see comments

annotator/pom.xml
ao508: i dont think we use an h2 database at all, this might be related to spring batch specifically. I think it creates one in memory to store job and execution context information?

annotationPipeline/src/main/java/org/cbioportal/annotation/AnnotationPipeline.java
inodb: Gotcha. Yeah that would be nice. Maybe we can do that as a follow up?

inodb:
Superseded by https://github.com/genome-nexus/genome-nexus-annotation-pipeline/pull/177

PR #167, opened 2021-02-17
Update spring version and cbioportal version

ao508:

### DIFF
--- a/annotationPipeline/src/main/resources/application.properties.EXAMPLE
+++ b/annotationPipeline/src/main/resources/application.properties.EXAMPLE
@@ -1,3 +1,4 @@
+spring.main.web-application-type=none
 spring.batch.job.enabled=false
 genomenexus.base=https://www.genomenexus.org
 genomenexus.isoform_query_parameter=isoformOverrideSource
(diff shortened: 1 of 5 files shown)


ao508:
@inodb Build is breaking because the annotationPipeline child module isn't recognizing its sibling module that's specified in its pom.... ideas? I'm thinking I'm just having tunnel vision and I'm just missing something obvious

Some other changes:
- added maven compiler version (not having this has burned us before)

inodb:
@ao508 maybe adding the spring boot configuration linked below will fix it? I vaguely recall seeing these errors. It seems limited to the annotator module

https://github.com/genome-nexus/genome-nexus-annotation-pipeline/pull/165/files#diff-c159c7b7adbb857d104c756c1b926200cb868aeaf18812adfc160048a336df83R48-R55

annotationPipeline/src/main/resources/application.properties.EXAMPLE
```diff
@@ -1,3 +1,4 @@
+spring.main.web-application-type=none
```
inodb, line 1: I still think we should handle this differently. Not the best user experience to have their pipeline crash after an upgrade. But ok to leave as is. Filed a separate issue for this: https://github.com/genome-nexus/genome-nexus-annotation-pipeline/issues/168

ao508: Will be spending more time on the build issue today so I'll take a look at this as well

PR #175, opened 2021-07-29
use cbioportal v3.7.0

inodb:
Rolling out new database version

PR #176, opened 2021-07-30
Handle logging changes for cbioportal 3.6.4 plus upgrade to 3.7.0

inodb:
This combines #165 with the logging changes + #175 the 3.7.0 upgrade. The standalone upgrade to 3.7.0 didn't work, because it needs the logging changes

The description of https://github.com/genome-nexus/genome-nexus-annotation-pipeline/pull/165 is better, so please see the discussion there for a better explanation of the changes

Note that there is a failing test, but it is already failing in master. We should address those in a separate PR. I'm guessing they are actually server issues since we get a status of annotation failed in the MAF

This supersedes #165 and #175

### DIFF
--- a/annotationPipeline/pom.xml
+++ b/annotationPipeline/pom.xml
@@ -58,4 +58,14 @@
     </dependency>
   </dependencies>
 
+  <build>
+    <plugins>
+      <plugin>
+        <!-- required to build an executable jar -->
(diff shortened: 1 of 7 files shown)


inodb:
superseded by https://github.com/genome-nexus/genome-nexus-annotation-pipeline/pull/177

PR #180, opened 2021-08-12
Add Dockerfile

SRodenburg:
A Dockerfile to call the GN annotation pipeline.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+FROM maven:3-openjdk-11
+
+ENV GN_HOME=/genome-nexus-annotation-pipeline
+COPY . $GN_HOME
+WORKDIR $GN_HOME
+
+RUN cp $GN_HOME/annotationPipeline/src/main/resources/application.properties.EXAMPLE $GN_HOME/annotationPipeline/src/main/resources/application.properties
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+# set log4j file in properties
+RUN sed -i "s|log4j\.appender\.a\.File.*|log4j.appender.a.File = $GN_HOME/genome-nexus-annotation-pipeline.log|" $GN_HOME/annotationPipeline/src/main/resources/log4j.properties
+
+ARG mvnprofiles=''
+RUN mvn -DskipTests clean install $mvnprofiles
+
+RUN mv $GN_HOME/annotationPipeline/target/annotationPipeline-*.jar $GN_HOME/annotationPipeline/target/annotationPipeline.jar
+
+ENTRYPOINT ["java", "-jar", "/genome-nexus-annotation-pipeline/annotationPipeline/target/annotationPipeline.jar"]
```
inodb, line 18: @SRodenburg should we do multi-stage building like we do for most other projects i.e. so the final `FROM` is just `openjdk11`?

inodb (approved):
Awesome! Thank you!

I made one minor comment, but we can merge as is if you prefer

SRodenburg:
@inodb should be ready to merge now. Not sure whether the tests should still be fixed?

SRodenburg:
@inodb should be ready to merge now. Not sure whether the tests should still be fixed?

Issue #193, opened 2022-06-09
Many output errors and warnings when running the minimal example


inodb:
Seems like things get annotated correctly but tons of errors about reading fields:

```
Something went wrong reading field Entrez_Gene_Id
Something went wrong reading field dbSNP_RS
Something went wrong reading field Match_Norm_Seq_Allele1
Something went wrong reading field Validation_Method
Something went wrong reading field Match_Norm_Seq_Allele2
Something went wrong reading field n_ref_count
Something went wrong reading field t_alt_count
Something went wrong reading field BAM_File
Something went wrong reading field Variant_Classification
Something went wrong reading field dbSNP_Val_Status
Something went wrong reading field Mutation_Status
Something went wrong reading field Matched_Norm_Sample_Barcode
Something went wrong reading field Validation_Status
Something went wrong reading field Variant_Type
Something went wrong reading field Strand
Something went wrong reading field Hugo_Symbol
Something went wrong reading field Sequencer
Something went wrong reading field n_alt_count
Something went wrong reading field Center
Something went wrong reading field Match_Norm_Validation_Allele2
Something went wrong reading field Tumor_Sample_Barcode
Something went wrong reading field Verification_Status
Something went wrong reading field t_ref_count
Something went wrong reading field Tumor_Seq_Allele2
Something went wrong reading field Match_Norm_Validation_Allele1
Something went wrong reading field Score
Something went wrong reading field Sequencing_Phase
Something went wrong reading field Tumor_Validation_Allele2
Something went wrong reading field Tumor_Validation_Allele1
Something went wrong reading field NCBI_Build
Something went wrong reading field Sequence_Source
Something went wrong reading field Entrez_Gene_Id
Something went wrong reading field dbSNP_RS
Something went wrong reading field Match_Norm_Seq_Allele1
Something went wrong reading field Validation_Method
Something went wrong reading field Match_Norm_Seq_Allele2
Something went wrong reading field n_ref_count
Something went wrong reading field t_alt_count
Something went wrong reading field BAM_File
Something went wrong reading field Variant_Classification
Something went wrong reading field dbSNP_Val_Status
Something went wrong reading field Mutation_Status
Something went wrong reading field Matched_Norm_Sample_Barcode
Something went wrong reading field Validation_Status
Something went wrong reading field Variant_Type
Something went wrong reading field Strand
Something went wrong reading field Hugo_Symbol
Something went wrong reading field Sequencer
[...]

PR #195, opened 2022-06-19
Unwanted logs while file loading

ozguzMete:
Fixes related to #193 
Logs are reserved for the devs.
Unnecessary merging removed.
Additional javadoc.


inodb (approved):
LGTM - not sure if we need bytebuddy

PR #196, opened 2022-06-19
Maf file validation

ozguzMete:
Checks hard requirements of the MAF file. In other words, it searches for these column names: Chromosome, Start_Position, End_Position, Reference_Allele. It then checks if either `Tumor_Seq_Allele1` or `Tumor_Seq_Allele2` are present


inodb (approved):
LGTM! Thank you!

inodb (changes requested):
@ozguzMete based on conversation here:

https://github.com/cBioPortal/cbioportal/pull/9656#discussion_r915100228

Can we actually change this to look for whether `Tumor_Seq_Allele1, `Tumor_Seq_Allele2` or both are present? Only if both are missing it should indicate that the input file misses a necessary field

ozguzMete:
@inodb updated the code and the tests accordingly.

Issue #194, opened 2022-06-14
Custom output format files


inodb:
Allow changing the column names etc using a supplied output format file

ozguzMete:
https://github.com/genome-nexus/genome-nexus-annotation-pipeline/pull/198

*magically* removing empty columns is much more difficult hence the data is row-based. Doable but painful... if the output format is enough for most of the time, we can skip this part

ozguzMete:
I have added _tcga_ and _minimal_ to the #198

PR #198, opened 2022-06-21
Output format

ozguzMete:
A possible solution for the issue #194

There are lots of empty columns in the output file. You will be able to use option-format option to supply an output format.
As an example, you can create a file with the following line:

Hugo_Symbol,Entrez_Gene_Id,Center,NCBI_Build,Chromosome

the application will only output these columns + any other column/s required by the application.
Order is preserved.

Currently, it only supports commas as the separator. You can use any number of spaces you like...


inodb (approved):
LGTM

rmadupuri (approved):
Tested and looks good! Thanks!

inodb:
Follow up as reported by @rmadupuri : https://github.com/genome-nexus/genome-nexus-annotation-pipeline/issues/210

PR #211, opened 2022-08-11
cleanup

ozguzMete:
Following cleanup operations are applied:

simplify regex

proofread

remove redundant groupId

remove redundant type cast

remove redundant array creation

remove redundant collection operations

simplify assertions

try-with-resources

fix package naming

try-with-resources

remove unused assignments

remove always true check

remove redundant addAll

use addAll

use List.sort

convert to lambda

merge identical catches

remove explicit types

fix illogical invocations

enhance the for loop

delete unnecessary imports

remove redundant exceptions

remove redundant local variable

remove redundant if

remove unnecessary conversion to string

remove unnecessary toString call

fix deprecated API usage

remove invalid hotspot code

fix obsolete assertions in JUnit5 tests


annotationPipeline/src/main/java/org/cbioportal/annotation/AnnotationPipeline.java
```diff
@@ -78,34 +78,35 @@ private static void help(Options gnuOptions, int exitStatus)
     }
 
     private static void launchJob(String[] args, String filename, String outputFilename, String outputFormat, String isoformOverride,
-            String errorReportLocation, boolean replace, Integer postIntervalSize) throws Exception
+            String errorReportLocation, boolean replace, String postIntervalSize) throws Exception
```
inodb, line 81: @ozguzMete should this really be a string? The interval size is a number i believe. Can't immediately see where it gets parsed to integer later

ozguzMete: @inodb it was used like this: .addString("postIntervalSize", String.valueOf(postIntervalSize)

so, we were converting to Integer then we were converting to String again (the line above), then Spring converts it back to integer inside of MutationRecordReader class using this: 

@Value("#{jobParameters[postIntervalSize]}")
    private Integer postIntervalSize; 

tests cover this class and I manually debugged to see assigned value after your post. No problem

inodb: ok perfect! Thank you!

PR #217, opened 2022-08-17
Fix redundant exception and help prints

ozguzMete:
With the addition of subcommands and throwing an exception when no subcommand is found, I started to see some redundant exception prints. Throw causes redundant help prints too.

This pull request fixes the issue


inodb (approved):
LGTM - thanks for the quick fix!

PR #221, opened 2022-08-22
Metering

ozguzMete:
I have added average response time to the summary

```
Annotation Summary:
	Records with ambiguous SNP and INDEL allele changes:  0

	Failed annotations summary:  3 total failed annotations
		Records with HGVSp null variant classification:  3
		Records that failed due to other unknown reason: 0

	Average Response Time:  0.010 sec.
```


inodb:
Thanks @ozguzMete ! Maybe total run time could be nice to add as well?

PR #226, opened 2022-09-04
Big File Handling

ozguzMete:
All the necessary changes to handle big files

**NOTE:** the code is experimental, and this PR shouldn't be approved


annotator/src/main/java/org/cbioportal/models/MutationRecord.java
```diff
@@ -28,500 +28,87 @@
...
-    public List<String> getHeaderWithAdditionalFields() {
-        List<String> headerWithAdditionalFields = new ArrayList<>(getHeader());
-        headerWithAdditionalFields.addAll(getSortedAdditionalPropertiesKeys());
-        return headerWithAdditionalFields;
+    // value is the header index
+    private static final Map<String, Integer> headers = new HashMap<>();
+    private static final String separator = "\t";
+    private final int[] beginIndices;
+    private final int[] endIndices;
+    private final String line;
```
sheridancbio, line 61: Here, rather than storing the fields separately we are storing a raw string which represents the record. It does not feel like we are modeling the data here .. because in this representation there are no assumptions about fields which are expected, must be present or may be optional, or any other semantics. This model could be used for any tab-delimited file format (like CNA or whatever) and it does not document to the programmer what the expectations are. With this code I could not come to the MutationRecord.java file in order to see what "belongs" in a MutationRecord easily. I'm not sure what else to say here ... this is big departure from our normal approach of having the model class be a representation of the expected elements of the object being modeled. We should probably discuss this.

annotator/src/main/java/org/cbioportal/models/MutationRecord.java
```diff
@@ -28,500 +28,87 @@
...
-        header.add("Validation_Method");
-        header.add("Score");
-        header.add("BAM_File");
-        header.add("Sequencer");
-        header.add("t_ref_count");
-        header.add("t_alt_count");
-        header.add("n_ref_count");
-        header.add("n_alt_count");
+    public Collection<String> getHeaders() {
+        return headers.keySet();
```
sheridancbio, line 111: same comment as above .. this seems to present the headers in an arbitrary order.

annotator/src/test/java/org/cbioportal/annotator/internal/GenomeNexusImplTest.java
```diff
@@ -47,415 +48,416 @@
...
-        map.put("SAMPLE-VARIANT-11", "4,77675978,77675979,-,C");
-        map.put("SAMPLE-VARIANT-12", "8,37696499,37696500,-,G");
-        map.put("SAMPLE-VARIANT-13", "10,101953779,101953779,T,-");
-        map.put("SAMPLE-VARIANT-14", "6,137519505,137519506,CT,-");
-        map.put("SAMPLE-VARIANT-15", "3,114058003,114058003,G,-");
-        map.put("SAMPLE-VARIANT-16", "9,135797242,135797242,C,AT");
-        map.put("SAMPLE-VARIANT-17", "6,137519505,137519506,CT,A");
-        return map;
-    }
+//    @Autowired
```
sheridancbio, line 61: This doesn't seem right ... we should not be commenting out the entire unit test class ... if we want to delete it we should delete it [but I would ask why it was being deleted].

annotator/src/test/java/org/cbioportal/models/MutationRecordTest.java
```diff
@@ -1,203 +1,36 @@
...
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.test.context.ContextConfiguration;
-import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
-//import org.springframework.test.util.ReflectionTestUtils;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 
-@ContextConfiguration(classes=MockGenomeNexusImpl.class)
-@RunWith(SpringJUnit4ClassRunner.class)
-public class MutationRecordTest {
+class MutationRecordTest {
```
sheridancbio, line 9: I have questioned the change of the MutationRecord model above. The semantic change so that the class represents an amorphous tab delimited record with no constraints on the fields present (or their data types / allowable values) would mean a drastic change to the unit tests obviously. Maybe the unit tests which were/are present here can point us at some of the constraints which we used to have and which are being dropped. I never did like the "additionalFields" part of our modeling. I thought that was mushy and unstructured. So getting rid of the weirdness of having an "everything else" set of values is a good thing. But I'm personally uncomfortable with the idea of making all the fields become additionalFields and having no core at all.

databaseAnnotator/pom.xml
```diff
@@ -1,122 +0,0 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
```
sheridancbio, line 1: I thought we chose to maintain the databaseAnnotator functionality ... Maybe @ao508 can comment.

sheridancbio (changes requested):
I've made a bunch of comments. Maybe my biggest critique is about changing the MutationRecord model into an abstract tab delimited field representation. I think that is something which should be discussed. Also I wonder if we have chosen to drop the database annotator now? About a year ago we used it and I thought we decided to continue to maintain it.
Also, after having finished the review, I think I missed the part which actually helped in the handling of large files to be annotated. I did see that batchsize increase. However I think that must be tested carefully against a deployment of VEP which is local (using genome-nexus-vep). I think that we tuned the request set size so that the local VEP servers would not be overwhelmed.
One nice thing I appreciate about the PR is all of the style / whitespace standardizations.

sheridancbio:
I just realized that this was part of the gsoc effort ... so gsoc mentors would be in a much better position to evaluate this PR than I am. I am happy to discuss my comments and to wait for mentors to lead the evaluation.

inodb:
We moved the essential bits for performance imporvement to https://github.com/genome-nexus/genome-nexus-annotation-pipeline/pull/228

inodb:
Thanks for reviewing @sheridancbio !

PR #227, opened 2022-09-09
Remove spring persistence and improve runtime by increasing default chunk size

ozguzMete:
- Removing spring persistence b/c it's not used
- By increasing default chunk size of the writer (so not the post body size), we are able to achieve ~50% runtime improvement

### DIFF
--- a/annotationPipeline/src/main/java/org/cbioportal/annotation/pipeline/BatchConfiguration.java
+++ b/annotationPipeline/src/main/java/org/cbioportal/annotation/pipeline/BatchConfiguration.java
@@ -32,8 +32,6 @@
 
 package org.cbioportal.annotation.pipeline;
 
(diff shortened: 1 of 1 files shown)


annotationPipeline/src/main/java/org/cbioportal/annotation/pipeline/BatchConfiguration.java
```diff
@@ -71,7 +59,7 @@
     @Autowired
     public StepBuilderFactory stepBuilderFactory;
 
-    @Value("${chunk:10}")
+    @Value("${chunk:1000000}")
```
inodb, line 62: note to self: this chunk size is related to the reader (so not post interval size)

inodb (approved):
Thanks so much!

PR #230, opened 2022-10-04
Remove cBioPortal Dependency

Luke-Sikina:
- Import MAFUtil from new dedicated MAF lib
- Remove cBioPortal import
- Import guava and commons directly rather than using the versions packaged in cBioPortal


averyniceday (approved):

averyniceday:
Can you see why the build is failing in circleci?

PR #231, opened 2022-10-11
Update documentation

leexgh:
Update README.md with additional fields and oncokb

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,89 @@ Make sure to adjust the file paths according to your specific requirements. Once
 | `-a` | `--add-original-genomic-location` | Add original genomic location data columns into the output, name columns with prefix 'IGNORE_Genome_Nexus_Original_'). This would be useful if saving a reference of original input is needed and won't be changed in any condition|
 | `-d` | `--ignore-original-location` | Genome-nexus-annotation-pipeline reads original genomic location info as input by default, if not existing, reading from normal genomic location info columns. Adding `-d` ignores original genomic location info columns (columns with prefix 'IGNORE_Genome_Nexus_Original_') and only use whatever in normal genomic location info columns. This would be helpful if you'd like to stick with current genomic location info columns.|
 
+### Annotation fields
+| Field | Source | Note |
+| --| -- | -- |
+| Hugo_Symbol |||
+|Entrez_Gene_Id|||
+|Center|||
+|NCBI_Build|||
+|Chromosome|GN response / Input||
+|Start_Position|GN response / Input||
+|End_Position|GN response / Input||
+|Strand|||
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -40,6 +40,89 @@ reporting to a file, supply the `-e` option a location for the file to be
...
+|MA:link.PDB|Mutation Assessor|Need to add "mutation_assessor" in "enrichment_fields"|
+|Polyphen_Prediction|Polyphen|Need to add "polyphen" in "enrichment_fields|
+|Polyphen_Score|Polyphen|Need to add "polyphen" in "enrichment_fields|
+|SIFT_Prediction|SIFT|Need to add "sift" in "enrichment_fields|
+|SIFT_Score|SIFT|Need to add "sift" in "enrichment_fields|
+|Ref_Tri||Need to add "nucleotide_context" in "enrichment_field|
+|Var_Tri||Need to add "nucleotide_context" in "enrichment_field|
+|Annotation_Status|||
+
+### Enrichment fields
```
inodb, line 112: Wondering if we should rephrase this header/section a bit. "Enrichment fields" is a genome nexus specific term, so maybe not super clear to a user? Another option is to say "Add additional annotation columns" or something along those lines and only mention that it involves setting `genomenexus.enrichment_fields=`

Some other points to consider here:

1. Good to indicate how to set this parameter on the command line for the compiled jar, rather than pre compilation phase. A user would prefer to use compiled source code rather than having to compile the entire project from source. Even better if there's a docker image, since they won't even need to install java. Many bioinformaticians would be familiar with how to use docker. See e.g. example here: https://docs.genomenexus.org/annotate-file
2. Since the term enrichment_fields is genome nexus specific it makes me wonder if it would be better for a user to eventually have a parameter like `--annotation-sources nucleotide_context,sift,mutation_assessor` etc. We could continue to use the term `enrichment_fields` in the code if we want, but for the user facing interfaces it might be better to use terminology that users are already familiar with

inodb (approved):
Thanks so much for figuring this out! This is great!

I made a few comments. Fine to address at a later stage

README.md
inodb: btw: for 1. something like this might work:

```

java
    -Dgenomenexus.enrichment_fields=annotation_summary,my_variant_info
    -jar annotationPipeline/target/annotationPipeline-*.jar \
    -r \
    --filename test/data/minimal_example.in.txt  \
    --output-filename test/data/minimal_example.out.uniprot.txt \
    --isoform-override uniprot
```
It would be good if we can make that a proper command line parameter for the java app itself (like `--annotation-source xxx`) rather than a system env variable

PR #236, opened 2022-12-08
Update to use new InfoController/Version Schema

averyniceday:
Update GNAP to use updated schema - (InfoControllerApi updated to ret…urn multiple versions such as GN server, database, VEP...)
Version outputted will be GN Server Version
Updated header in expected outputs for integration tests

There are four tests that are still failing -- those are independent of this PR. I tested this by manually setting the returned version in unchanged code and running tests. 

https://github.com/genome-nexus/genome-nexus-java-api-client/pull/14

DO NOT MERGE until PR in genome-nexus-java-api-client is updated


leexgh (commented):
Thanks! Looks good to me.

Here are what I found about failing tests:

I also noticed the failing test: `check_if_my_variant_info_provides_gnomad_annotations` recently. It's because my_variant_info changed their API model so we couldn't map their response back to our model. We could do a temporary fix to update our model, but the fix couldn't last long, so probably the best way to fix is to host gnomAD data in GN database instead of getting it from my_variant_info. Not sure if we want to hold until importing the data or have a temp fix for the failing test.

`check_if_maf_file_still_the_same_when_annotating_with_uniprot_transcripts` and `check_if_nucleotide_context_provides_Ref_Tri_and_Var_Tri_columns` look like related to the update of `common_input/isoform_overrides_uniprot.txt` in [this PR](https://github.com/genome-nexus/genome-nexus-importer/pull/54), which uses the latest uniprot mapping from [vcf2maf repo](https://github.com/mskcc/vcf2maf/blob/main/data/isoform_overrides_uniprot). 

Similar to `check_if_maf_file_still_the_same_when_annotating_with_mskcc_transcripts` which is related to the updating of `common_input/isoform_overrides_mskcc_grch37.txt` from [vcf2maf](https://github.com/mskcc/vcf2maf/blob/main/data/isoform_overrides_at_mskcc).

Besides the test files, do we need to update "genome_nexus_version" in [example files](https://github.com/genome-nexus/genome-nexus-annotation-pipeline/tree/master/test/data)? How do we keep the version number in sync with the latest deployment in the future?

PR #241, opened 2023-02-23
Add mvn-external-version

inodb:
This updates the pom version based on the git version/commit. This is the same as it works for cBioPortal:
https://github.com/cBioPortal/cbioportal/pull/5403.


leexgh (approved):
👍

PR #242, opened 2023-02-23
Add client version as subcommand

inodb:
Adds a command like:

```
java -jar annotationPipeline/target/annotationPipeline-*.jar version
Client: 0f6a7b8-dirty-SNAPSHOT
```

Idea is to extend with server version etc as well at some point

### DIFF
--- a/annotationPipeline/src/main/java/org/cbioportal/annotation/AnnotationPipeline.java
+++ b/annotationPipeline/src/main/java/org/cbioportal/annotation/AnnotationPipeline.java
@@ -55,6 +55,7 @@ import java.time.Duration;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Properties;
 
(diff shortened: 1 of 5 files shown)


annotationPipeline/src/main/java/org/cbioportal/annotation/AnnotationPipeline.java
```diff
@@ -114,20 +115,43 @@ public static void subMain(String[] args) throws NoSubcommandFoundException, Par
...
+        java.util.Properties p = new Properties();
+        try {
+            p.load(is);
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+        String appVersion = p.getProperty("app.version");
+
+        return appVersion;
+    }
```
inodb, line 147: wasn't able to get this working with standard spring decorators, because it's all using static methods etc

PR #250, opened 2023-07-31
Change ENTRYPOINT to CMD for run flexibility

hweej:
Nextflow has deprecated support for executable containers. e.g ENTRYPOINT defaults.
[529](https://github.com/nextflow-io/nextflow/issues/529) 

Changing ENTRYPOINT to CMD would keep the default behavior of running annotationPipeline.jar. It also has the added benefit of possibly pulling in [annotation-tools](https://github.com/genome-nexus/annotation-tools/tree/master)


inodb (approved):
Thanks! LGTM

PR #252, opened 2023-07-31
Add support for oncokb annotation

sivkovic:
Enable annotation with OncoKB. To add Oncokb annotation, provide oncokb token in `application.properties`, (e.g. oncokb.token=ewriufh234hisd23798fskj)


inodb (approved):
LGTM - thank you!

PR #253, opened 2023-07-31
Add procps for ps command

hweej:
Nextflow workflow tracing requires the ps tool to be installed inside the docker container


inodb (approved):
LGTM - thanks!

PR #256, opened 2023-08-16
Fix unsupported enrichment fields in query

leexgh:
Fix: https://github.com/genome-nexus/genome-nexus-annotation-pipeline/issues/257
`sift` and `polyphen` are not supported in genome nexus query "fields", but it's available as one of the "enrichment_fields" in annotation pipeline. Because the annotation of `sift` and `polyphen` are resolved from canonical transcript, which is from VEP, so no need to pass `sift` and `polyphen` to genome nexus server. Adding `sift` and `polyphen` in "enrichment_fields" 
 only serves as a flag to add those columns to the output annotation file.


inodb (approved):
LGTM! Thank you!!

Issue #255, opened 2023-08-15
Add note column to include some important comments


leexgh:
**Background**:
- If there are common prefix being removed, or end position is altered (when end position is missing or is not correct), start position and / or end position might be updated. We have a new field in genome nexus response, e.g. `"genomicLocationExplanation":"End position changes from 123 to 122, because end position should be the position of last deleted nucleotide.`
- Add comments for failing variants, we have a new field called `errorMessage`, e.g. `  "errorMessage": "Unable to parse HGVS notation '12:g.58142319A>C': Reference allele extracted from 12:58142319-58142319 (T) does not match reference allele given by HGVS notation 12:g.58142319A>C (A)",
`
    - wrong coordinates, alleles
    - no matches in VEP
 - server issues

**Proposal**:
- Add new parameter for instance `-n` `--note`.
- If have `-n`, add a new column:`GenomicLocation Explanation`.
- (if user defines `-e`) In error report, add `Error Message` column using the "errorMessage" value.

PR #265, opened 2023-09-14
Add '-n' note column in maf and error message into error report

leexgh:
Fix: https://github.com/genome-nexus/genome-nexus-annotation-pipeline/issues/255
- new parameter `-n`
- `-n` will add "genomic_location_explanation" column to output file. Only variants that have altered genomic location will have value in this column 
    - example column value: `End position changes from 170837525 to 170837526, end position should equal to start position for SNV variants`
- Add error message in error report. To create error report, add `-e REPORT_PATH` in command. 
    - Example error message: `Reference allele extracted from response (CA) does not match given reference allele (AC).`
- Add error message into LOG file. 
    - Example log: `2023-09-13 22:45:41 [main] WARN  org.cbioportal.annotator.internal.GenomeNexusImpl - Annotation failed for variant 17:g.7578475del;Reference allele extracted from response (TGCCGGGCG) does not match given reference allele (GACGCGGGT)`
- Update documentation about `-n`


rmadupuri (approved):

leexgh:
@rmadupuri Not sure if we want to add `-n` by default, for most of the variants this column will be empty, only variants with altered genomic location will have value, so I'm not sure if it makes sense to do it for all variants. 
Have you ever seen a column reach the limit? Do you know what is the limit?

README.md
```diff
@@ -71,11 +67,12 @@ Make sure to adjust the file paths according to your specific requirements. Once
...
 | `-d` | `--ignore-original-location` | Genome-nexus-annotation-pipeline reads original genomic location info as input by default, if not existing, reading from normal genomic location info columns. Adding `-d` ignores original genomic location info columns (columns with prefix 'IGNORE_Genome_Nexus_Original_') and only use whatever in normal genomic location info columns. This would be helpful if you'd like to stick with current genomic location info columns.|
+| `-n` | `--not-column` | The `-n` command line parameter provides information about genomic location corrections that may occur in Genome Nexus. In certain corner cases, the end position of a genomic location might be missing or incorrect. When you use this parameter, a new column named "genomic_location_explanation" will be added to the annotation output. If alterations occur, the "genomic_location_explanation" column will provide detailed information about why and how the genomic location was modified. If you prefer to keep a copy of your original genomic location, considering add `-a` in your command.
```
inodb, line 75: this should maybe be:

`--harmonization-note-column`?

PR #270, opened 2024-01-09
Return non-zero exit code when annotation pipeline fails

jamesqo:
This PR adds a reliable way of checking whether or not the Genome Nexus pipeline failed from the command line. It is needed for the NCI work because we want to distinguish the cases where it is an annotation issue (eg. missing variants) or with the GN code itself.

/cc @averyniceday @leexgh @inodb


leexgh (approved):
Thanks for adding this!

PR #272, opened 2024-01-25
Upgrade to java 21

mandawilson:
Changes needed for running under java21 and for packaging this depenency into the pipelines codebase for cbioportal study import into msk databases.


.circleci/config.yml
```diff
@@ -32,7 +32,7 @@ jobs:
 
       - run:
           name: "compile"
-          command: "mvn clean install"
+          command: "mvn clean install -Dnet.bytebuddy.experimental=true"
```
inodb, line 35: what is the experimental thing for?

mandawilson: One of the dependencies (bytebuddy from Mockito) won't work with Java 21 without it.  But let me try to update Mockito and see if that fixes it.

mandawilson: Phew!  😅 It wasn't easy but I finally got the right set of dependencies so we don't need that flag.  I think the "-Dnet.bytebuddy.experimental=true" was temporarily needed for Java 21 until recently.

sheridancbio (approved):

leexgh (commented):
Thanks for the updates! Do we need to update Java version in docker as well?

pom.xml
```diff
@@ -77,21 +78,6 @@
...
-        <exclusion>
-          <groupId>ch.qos.logback</groupId>
-          <artifactId>logback-core</artifactId>
-        </exclusion>
-        <exclusion>
-          <groupId>ch.qos.logback</groupId>
-          <artifactId>logback-classic</artifactId>
-        </exclusion>
-      </exclusions>
-    </dependency>
```
inodb, line 94: @sheridancbio do we still want to keep this?

sheridancbio: Yes, for now I think we should keep this dependency. @haynescd pointed out that maf is a dependency in cbioportal-core, and since the maf code is used in both this repository and in that one we can keep it in a separate repository I think.

Hopefully we can develop a unified community-wide tool for study import which includes the ability to annotate variants (via genome nexus). That would essentially replace cbioportal-core. At that time we could decide whether the ability to parse and represent variants from a mutation file should be its own module or become embedded in one of these two repositories (most likely this one).

sheridancbio: Note that we still do have the dependency .. it is simply moved to the annotator module (it is only needed there)

inodb:
@mandawilson @sheridancbio i noticed there's an error in the build here - are you still looking at this one? Thanks!

sheridancbio:
> @mandawilson @sheridancbio i noticed there's an error in the build here - are you still looking at this one? Thanks!

Yes .. we are trying to resolve conflicts with the pipelines codebase (which also packages in cbioportal-core repository. There are spring version differences between these two codebases so we may need to make the spring version of this codebase match the spring version (and boot version) in use in cbioportal-core in order for the importer to function correctly.

So consider this PR to be in flux for now.

sheridancbio:
This PR seems complete now. Docker functionality has been tested. Note: the "merge" sub command mode (seen through docker build and run) is not running either in this PR or in the master branch. A new issue to repair (or delete) the merge code logic should be created (if not extant).

Studies have been annotated during import in the msk importer (which uses the annotator module as a component).

changes sets were squashed into a single changeset.

sheridancbio:
Note: the merge subcommand functionality actually did succeed when executed with the proper arguments ... basically those were: java -jar {jar_filename} merge --input-mafs-list='fileA,fileB,fileC,...' --output-maf='outfilename.txt'

I am proceeding with this merge now. A description of the potential ramifications was presented to the development team in the `genomenexus` slack channel in the cbioportal organization and there were no objections. This included the dropping of the databaseAnnotator module, the new requirement to build and run this codebase under a java21 jdk, and the successful testing of the annotator module as a dependency in the msk importer and also in the Docker mode execution for annotating test files.

PR #282, opened 2024-06-25
Add consequence for IGR variants

leexgh:
Part of: https://github.com/genome-nexus/genome-nexus/issues/725
Background: when importing data into cbioportal, we filter the variants by variant_classification and consequence. But intergenic variants have no value in these columns, so those variants will be imported and shown as `MUTATED`. We'd like to add `IGV` and `intergenic_variant` into annotated file, so those variants can be filtered out correctly.

Changes:
For intergenic variants, this pr adds`intergenic_variant` and `IGV` in `Consequence` and `Variant_Classification` columns, instead of having blank columns.
Previously:
image

New:
image


annotator/src/main/java/org/cbioportal/annotator/util/AnnotationUtil.java
```diff
@@ -331,11 +331,18 @@ public String resolveHgvspShort(TranscriptConsequenceSummary canonicalTranscript
...
-        if (canonicalTranscript == null || canonicalTranscript.getConsequenceTerms() == null) {
-            return "";
-        } else {
+    public String resolveConsequence(VariantAnnotation gnResponse, TranscriptConsequenceSummary canonicalTranscript) {
+        if (canonicalTranscript != null && canonicalTranscript.getConsequenceTerms() != null) {
             return canonicalTranscript.getConsequenceTerms();
+        } else if (gnResponse != null && 
+                   gnResponse.getAnnotationSummary() != null && 
+                   gnResponse.getAnnotationSummary().getIntergenicConsequenceSummaries() != null &&
+                   !gnResponse.getAnnotationSummary().getIntergenicConsequenceSummaries().isEmpty()) {
```
onursumer, line 340: Should we move this to a method like `hasIntergenicConsequenceSummaries` and reuse in the other method as well?

onursumer (approved):
LGTM!

annotator/src/main/java/org/cbioportal/annotator/util/AnnotationUtil.java
leexgh: Moved it out to a separate util funcion

PR #284, opened 2024-08-09
add alphaMissense Annotation

JunheZoooo:
# Add AlphaMissense Pathogenicity Prediction and Score to Genome Nexus Annotation Pipeline

## Description

This Pull Request adds two new columns, `alphamissense_pathogenicity` and `alphamissense_class`, to the Genome Nexus annotation pipeline. These columns will be included in the output file when AlphaMissense data is available.

Screenshot 2024-08-09 at 1 49 20 PM


rhshah:
Just a thought @JunheZoooo, VEP has https://github.com/Ensembl/VEP_plugins/blob/release/112/AlphaMissense.pm 
@inodb  can we confirm if VEP Plugins can be used with VEP used in Genome Nexus.

leexgh:
@rhshah We have alpha missense in genome nexus now, but it's only available for a few variants for testing purpose(see example: [7:g.140453136A>T](https://www.genomenexus.org/annotation/7:g.140453136A%3ET?fields=annotation_summary)), we still need to come up with a plan to support this for all variants and also maintaining the consistency of vep annotation

PR #285, opened 2024-10-09
Add mutation assessor v4 columns

leexgh:
Part of: https://github.com/genome-nexus/genome-nexus/issues/753
Removed columns that are not available in new version, add three new columns: `MSA`, `MAV`, `SV`
Example values:
image


onursumer (approved):
LGTM!

PR #286, opened 2025-02-20
Update document to use mskcc as the default isoform override

leexgh:
Partial fix: https://github.com/genome-nexus/genome-nexus/issues/626
Update document to recommend using mskcc as the isoform override source.

### DIFF
--- a/CMD_HELP.md
+++ b/CMD_HELP.md
@@ -18,7 +18,7 @@ annotate subcommand options:
  -e,--error-report-location <arg>   Error report filename (including path)
  -f,--filename <arg>                Mutation filename
  -h,--help                          shows this help document and quits.
- -i,--isoform-override <arg>        Isoform Overrides (mskcc or uniprot)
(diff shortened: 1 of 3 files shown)


CMD_HELP.md
```diff
@@ -18,7 +18,7 @@ annotate subcommand options:
  -e,--error-report-location <arg>   Error report filename (including path)
  -f,--filename <arg>                Mutation filename
  -h,--help                          shows this help document and quits.
- -i,--isoform-override <arg>        Isoform Overrides (mskcc or uniprot)
+ -i,--isoform-override <arg>        Isoform Overrides. Options: `mskcc` (preferred) or uniprot (legacy)
```
leexgh, line 21: @inodb Not sure if we should call uniprot "legacy"?

inodb: I like that idea yea 👍

inodb (approved):

PR #288, opened 2025-05-16
Read and process mutation record in batches to reduce memory usage

leexgh:
Fix: https://github.com/genome-nexus/genome-nexus/issues/796
**Problem**
`gnResponseVariantKeyMap` grows with `VariantAnnotation` added for each variant. It consumes a lot of memory when annotating a large file.
**Updates**
1. This pr tries to process and write in smaller batches to avoid high memory usage.
2. Updates for Annotation Summary. Previously, the annotation summary of invalid or failed annotations was a count for unique variants, now it shows the actual variant count, including duplicated variants. For example:
```
Annotation Summary:
        Records with ambiguous SNP and INDEL allele changes:  0

        Failed annotations summary:  5 total failed annotations
                Records with HGVSp null variant classification:  1
                Records that failed due to other unknown reason: 4
```
This makes more sense to me because when annotating a large file, usually people don't know how many variants are unique, including duplicated variants makes it clearer to tell people how many rows have invalid annotations.

### DIFF
--- a/annotator/src/main/java/org/cbioportal/annotator/internal/AnnotationSummaryStatistics.java
+++ b/annotator/src/main/java/org/cbioportal/annotator/internal/AnnotationSummaryStatistics.java
@@ -32,14 +32,17 @@
 
 package org.cbioportal.annotator.internal;
 
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import org.apache.commons.lang.StringUtils;
 import org.cbioportal.annotator.Annotator;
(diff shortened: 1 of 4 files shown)


leexgh:
@inodb @onursumer @ao508 @rmadupuri @callachennault Not sure who should review this pr, feel free to leave comments or unassign yourself from review 🙂

ao508 (commented):
lgtm but I'd like someone from the pipelines team to confirm that these changes will not impact the ETLs. Looks like the POST method is the same so when the annotator dependency is updated in say, the DMP data fetcher, it should be fine @callachennault

PR #293, opened 2025-08-07
Update Dockerfile to use maven 3.9.9

leexgh:
Fix: https://github.com/genome-nexus/genome-nexus/issues/800
Since mvn [3.9.10](https://maven.apache.org/docs/3.9.10/release-notes.html), it updates the Resolver engine version. Maybe concurrent builds are changed or verification logic is changed. Dynamic versioning is misaligned for parent and child since this version. However 3.9.9 works (from 2 months ago), switching to use the static version of mvn 3.9.9


inodb (approved):
LGTM – thank you!

